PR #5, opened 2020-11-18
Create a provider package

### LINKED ISSUE #1: How to install this provider?
This seems to be a nice project, but I struggle with installing the provider for testing. Could you give me some hints on how to build/install this provider into a crossplane installation?

srueg:
Closes #1

### DIFF
--- a/package/crossplane.yaml
+++ b/package/crossplane.yaml
@@ -1,0 +1,10 @@
+apiVersion: meta.pkg.crossplane.io/v1alpha1
+kind: Provider
+metadata:
+  name: provider-sql
+  annotations:
+    descriptionShort: |
+      The SQL Crossplane provider enables management of SQL servers.
+spec:
+  controller:
+    image: crossplane/provider-sql-controller:VERSION


negz (approved):

srueg:
Will this and #6 be enough to publish the package & controller images?

I guess the `DOCKER_USR`, `DOCKER_PSW` and `AWS_USR`, `AWS_PSW` secretes need to be setup in GH actions? Is the idea to already publish the artifacts or do you want to wait for more functionality?

negz:
> Will this and #6 be enough to publish the package & controller images?

I think so, off the top of my head. I believe those secrets are set at the org level for crossplane-contrib. I'm okay publishing artifacts; they should be pushed as `:master` tags iirc until we create release branches.

srueg:
Looks like the "Login to Docker" and "Publish Artifacts" steps are being skipped: https://github.com/crossplane-contrib/provider-sql/runs/1417312119?check_suite_focus=true

PR #6, opened 2020-11-18
Add GitHub Actions Setup

srueg:
As copied from https://github.com/crossplane-contrib/provider-helm


srueg:
Applied version from https://github.com/crossplane/crossplane

negz (approved):

Issue #3, opened 2020-11-17
Provision MySQL Grants


srueg:
Implement support for provisioning MySQL [grants](https://mariadb.com/kb/en/grant/) for users. In addition to provisioning databases and users on SQL servers grants should also be supported. We'll start with the implementation for MySQL. Support for PostgreSQL can follow later.

## Implementation Details
* Create a new manged resource `Grant` in `mysql.sql.crossplane.io/v1alpha1` with the following fields:
  * `.spec.forProvider.rawGrants` string array of raw `GRANT` statements
  * `spec.forProvider.grants` array with typed grants:
    * `.userRef` reference to a `User` resource (see #2) to which to apply the grant (e.g. `GRANT ... TO username`)
    * `.databaseRef` reference to a `Database` resource to which to apply the grant (e.g. `GRANT ... ON db_name.* ...`)
    * `.privileges` array of [privileges](https://mariadb.com/kb/en/grant/#database-privileges) to grant (e.g. `GRANT ... ALL ...`)
* Implement a new `grant` reconciler in the `pkg/controller/mysql/grant` package
* Provision a grant based on the CRD

## To be Discussed
How much logic should be provided to create grants?
I think the approach with both a `rawGrant` to define arbitrary statements and the typed `grants` fields would allow a simple solution for the most basic use cases and a more advanced way for fine granular permissions.

negz:
@srueg What do you think about something like this?

```yaml
apiVersion: mysql.sql.crossplane.io/v1alpha1
kind: Grant
metadata:
  # ...
spec:
  # A plain string - this user doesn't need to be known to Crossplane.
  user: foo
  # A cross resource reference to a MySQL "User" managed resource.
  # Used to set 'user' above.
  # userRef:
  #   name: foouser
  # A plain string - this databse doesn't need to be known to Crossplane.
  database: foo
  # A cross resource reference to a MySQL "Database" managed resource.
  # Used to set 'database' above.
  # databaseRef:
  #   name: foodb
  privileges: ['some', 'privs']
```

The key points here are:

* A `Grant` represents at most one grant, not many.
* We allow the author to grant users access to databases even if the user and/or database are not managed by Crossplane (which I presume was the goal of the `rawGrants` approach).

srueg:
Nice, I'll go with this approach 👌 
During implementation I also realized that multiple grants in one resource make things much more complicated...

[...]

PR #7, opened 2020-11-18
MySQL User

### LINKED ISSUE #2: Provision MySQL Users
Implement support for provisioning MySQL users. In addition to provisioning databases on SQL servers users should also be supported. We'll start with the implementation for MySQL. Support for PostgreSQL can follow later.

## Implementation Details
* Create a new manged resource `User` in `mysql.sql.crossplane.io/v1alpha1`
* Implement a new `user` reconciler in the `pkg/controller/mysql/user` package
* Provision a user based on the CRD
  * The external name of the managed resource should be used as the user name
  * A password should be generated
  * The complete [ConnectionDetails](https://pkg.go.dev/github.com/crossplane/crossplane-runtime@v0.11.0/pkg/reconciler/managed#ConnectionDetails) should be returned

## Not in Scope
* User TLS options
* User password expire options
* User resource options
* User GRANTs (see #3)

## To be Discussed
I think a more ergonomic (i.e. typed) way to define GRANTs might make sense in the future. For now I'd like to start with a simple approach.

srueg:

Fixes #2 

Based on the refactoring in https://github.com/crossplane-contrib/provider-sql/pull/8

I have:

Manually on a Kind cluster, in the unit tests and with an integration test.

`make test`
`make e2e.run`


apis/mysql/v1alpha1/databse_types.go
```diff
@@ -0,0 +1,58 @@
+/*
```
negz, line 1: It looks like there's a typo in this filename - `databse` rather than `database`.

srueg: Good catch, thanks 👍

srueg:
The basics are implemented. Some refactoring was necessary of the existing code base, see #8.
What I don't quite understand is how the password could be updated/rotated? Should we add a secretRef to define a password instead of generating one and track its `resourceVersion`?

negz:
> Should we add a secretRef to define a password instead of generating one and track its resourceVersion?

That sounds like a good idea; I believe we've started using that pattern in a few other places (RDS instances perhaps?).

negz (commented):
Whoops - I started a review yesterday but forgot to submit. Looks like some of my comments might be updated now, but I'll push anyhow as I don't have time to review fully today.

pkg/clients/sql/postgresql.go
```diff
@@ -0,0 +1,76 @@
...
+// IsDoesNotExist returns true if the supplied error indicates a database does
+// not exist.
+func (c PostgresDB) IsDoesNotExist(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// TODO(negz): Is there a less lame way to determine this?
+	return strings.HasSuffix(err.Error(), "does not exist")
+}
```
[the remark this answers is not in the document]
srueg: Even better: I got rid of it altogether since the `IF EXISTS` query handles this already.

srueg:
Thanks a lot for your review, really appreciate it! ✌️

negz (approved):
Thanks for working on this @srueg - it looks great! It may be worth updating the README either as part of this PR or a future one to document that we now support MySQL users.

srueg:
Thanks again for the review! I'll give #3 a shot next.
Do you agree with the proposed design of the API?

PR #8, opened 2020-11-18
Refactor Reconcilers

srueg:

Refactor the reconcilers and introduce a dedicated `clients` package.

I have:

Manually on a Kind cluster

### DIFF
--- a/pkg/clients/sql/common.go
+++ b/pkg/clients/sql/common.go
@@ -1,0 +1,26 @@
+package sql
+
+import (
+	"context"
+	"errors"
+
+	"database/sql"
+)
+
+// A Query that may be run against a DB.
+type Query struct {
+	String     string
+	Parameters []interface{}
+}
+
+// A DB client.
+type DB interface {
+	Exec(ctx context.Context, q Query) error
+	Scan(ctx context.Context, q Query, dest ...interface{}) error
+	IsDoesNotExist(err error) bool
+}
+
+// IsNoRows returns true if the supplied error indicates no rows were returned.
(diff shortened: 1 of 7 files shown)


pkg/controller/constants/errors.go
```diff
@@ -0,0 +1,12 @@
...
+const (
+	// ErrTrackPCUsage indicates tracking errors
+	ErrTrackPCUsage = "cannot track ProviderConfig usage"
+	// ErrGetPC indicates inaccessible ProviderConfig
+	ErrGetPC = "cannot get ProviderConfig"
+	// ErrNoSecretRef indicates inexistent SecretRef
+	ErrNoSecretRef = "ProviderConfig does not reference a credentials Secret"
+	// ErrGetSecret indicates inaccessible secret
+	ErrGetSecret = "cannot get credentials Secret"
+)
```
negz, line 12: This would be a new convention for Crossplane; we typically expose error strings as private constants in the same package they're used to make it easy to compare errors in tests. What is the advantage of creating a dedicated package for them?

srueg: The idea was to share the common error strings between the reconcilers instead of duplicating them. But I don't have a strong preference on it.

negz: I slightly prefer to maintain the status quo here, but I also don't feel strongly. :)

negz (approved):

pkg/controller/constants/errors.go
srueg: I changed it back 👍

PR #12, opened 2020-11-23
Provision MySQL Grants

### LINKED ISSUE #3: Provision MySQL Grants

srueg:

Fixes #3 

I have:

* Manually against Kind cluster
* `make test`
* `make e2e.run`


srueg:
Ok, I think this is ready for a review.
Unfortunately I couldn't find a better way to get current GRANTS than regex.
Also, when updating the GRANTS for a user, `REVOKE ALL` is first called and then the current GRANTS are applied.

I'll write some unit tests next.

apis/mysql/v1alpha1/grant_types.go
```diff
@@ -0,0 +1,135 @@
...
+// +kubebuilder:object:root=true
+
+// GrantList contains a list of Grant
+type GrantList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Grant `json:"items"`
+}
+
+// GetUsername returns the referenced username
```
negz, line 97: We actually have a library to help implement this pattern since it's fairly common. Sorry I neglected to point this out earlier! If your type satisfies the `ReferenceResolver` interface we'll automatically resolve references. You can use `pkg/reference` from crossplane-runtime to implement that interface in a fashion pretty similar to what you have here.

https://github.com/crossplane/crossplane-runtime/blob/57ef78/pkg/reconciler/managed/reconciler.go#L147
https://github.com/crossplane/provider-aws/blob/67fbb0c1e/apis/database/v1beta1/referencers.go

srueg: Wow, this is awesome. Makes it a lot easier.
Is it safe to just dereference the `Database` and `User` pointers without `nil` checks (since they should always be set by the reference resolvers, right)?

srueg: >  Is it safe to just dereference the `Database` and `User` pointers without `nil` checks (since they should always be set by the reference resolvers, right)?

Ah, I see. When I configure my reconciler with `managed.WithReferenceResolver(managed.NewAPISimpleReferenceResolver(mgr.GetClient())),` then the pointer should always be set.

srueg:
Ping @negz for another review, please

pkg/controller/mysql/database/reconciler.go
```diff
@@ -56,7 +56,8 @@ func Setup(mgr ctrl.Manager, l logging.Logger) error {
 	t := resource.NewProviderConfigUsageTracker(mgr.GetClient(), &v1alpha1.ProviderConfigUsage{})
 	r := managed.NewReconciler(mgr,
 		resource.ManagedKind(v1alpha1.DatabaseGroupVersionKind),
-		managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), usage: t, newDB: xsql.NewMySQLDB}),
+		managed.WithInitializers(managed.NewDefaultProviderConfig(mgr.GetClient()), managed.NewNameAsExternalName(mgr.GetClient())),
```
negz, line 59: I think this is equivalent to the default initializers - so omitting `WithInitializers` should have the same result.

pkg/controller/mysql/grant/reconciler.go
```diff
@@ -0,0 +1,324 @@
...
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	"github.com/crossplane-contrib/provider-sql/apis/mysql/v1alpha1"
+	"github.com/crossplane-contrib/provider-sql/pkg/clients/mysql"
+	"github.com/crossplane-contrib/provider-sql/pkg/clients/xsql"
+	mysqldriver "github.com/go-sql-driver/mysql"
```
negz, line 41: Nit: This should be up with the third party imports. (Our linter seems to be inconsistent about enforcing this rule.)

pkg/controller/mysql/grant/reconciler.go
```diff
@@ -0,0 +1,324 @@
...
+
+	privileges, result, err := c.getPrivileges(ctx, username, dbname)
+	if err != nil {
+		return managed.ExternalObservation{}, err
+	}
+	if result != nil {
+		return *result, nil
+	}
+
+	if !comparePrivileges(cr.Spec.ForProvider.Privileges, privileges) {
```
negz, line 147: Naming nit: Maybe something like `!equal` or `!privilegesEqual` rather than `!comparePrivileges` would more clearly indicate that this returns true if the privileges are equal.

pkg/controller/mysql/grant/reconciler.go
```diff
@@ -0,0 +1,324 @@
...
+
+	var privileges []string
+	for rows.Next() {
+		var grant string
+		if err := rows.Scan(&grant); err != nil {
+			return nil, nil, errors.Wrap(err, errCurrentGrant)
+		}
+		p := parseGrant(grant, dbname)
+		if p != nil && privileges != nil {
+			// Found more than one grant for this user on this DB
```
negz, line 205: Do we support creating multiple Grant resources for one User and Database? Should we?

pkg/controller/mysql/grant/reconciler.go
```diff
@@ -0,0 +1,324 @@
...
+				ResourceExists:   true,
+				ResourceUpToDate: false,
+			}, nil
+		}
+		privileges = p
+	}
+	if err := rows.Err(); err != nil {
+		var myErr *mysqldriver.MySQLError
+		if errors.As(err, &myErr) {
+			if myErr.Number == errCodeNoSuchGrant {
```
negz, line 216: Nit: Could this be && rather than a nested if?

pkg/controller/mysql/grant/reconciler.go
```diff
@@ -0,0 +1,324 @@
...
+	query := createGrantQuery(privileges, dbname, username)
+	if err := c.db.Exec(ctx, xsql.Query{String: query}); err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreateGrant)
+	}
+	if err := c.db.Exec(ctx, xsql.Query{
+		String: "FLUSH PRIVILEGES",
+	}); err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errFlushPriv)
+	}
+	return managed.ExternalCreation{}, nil
```
negz, line 249: Nit: `errors.Wrap` will return `nil` if `err` is `nil`, so you can just write:

```go
err := c.db....
return managed.ExternalCreation{}, errors.Wrap(err, errFlushPriv)
```

Ditto for `Update` and `Delete`.

pkg/controller/mysql/grant/reconciler.go
```diff
@@ -0,0 +1,324 @@
...
+
+	privileges := strings.Join(cr.Spec.ForProvider.Privileges, ", ")
+	username, host := mysql.SplitUserHost(username)
+
+	// Remove current grants
+	query := fmt.Sprintf("REVOKE ALL ON %s.* FROM %s@%s",
+		mysql.QuoteIdentifier(dbname),
+		mysql.QuoteValue(username),
+		mysql.QuoteValue(host),
+	)
```
negz, line 269: Could this result in applications running as this user being unable to use the database? e.g. For a small period of time between this query being executed and the following create query being executed? Or perhaps even longer if this query succeeds but the create fails? Is it possible to do this in a transaction?

srueg: Yes, this could lead to applications having no privileges for a short time. I couldn't find a way to `UPDATE` a grant and not removing it first...

Unfortunately, `GRANT` and/or `REVOKE` statements can't be used in transactions since they lead to an implicit commit of said transaction: https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html

pkg/controller/mysql/grant/reconciler.go
srueg: Currently only a single `Grant` resource can exist for every user/database combination. Multiple privileges on a single Grant are possible though.
This place in the code basically assumes some external system changed the grants and we should update them again according to the controller's state of truth.

I don't see a way to support multiple Grant resources for the same user/database combination since we couldn't differentiate which MySQL grant is coming from which `Grant` resource. Also as far as I saw, MySQL combines the privileges of multiple `GRANT` queries for the same user/database combo.

pkg/controller/mysql/grant/reconciler.go
negz: It might be worth capturing this information in a comment for future readers.

pkg/controller/mysql/grant/reconciler.go
negz: Got it. It might be a good idea to explain this in a comment - it could be obvious to those more versed in SQL but it was not to me.

negz (approved):
Thanks @srueg! I'm excited to have more complete support in this provider.

PR #19, opened 2021-01-07
Postgresql Role support

### LINKED ISSUE #17: Implement PostgreSQL Role management
Lack of Postgres User management.

Note: `CREATE USER` in recent Postgres versions is an alias for [`CREATE ROLE WITH LOGIN`](https://www.postgresql.org/docs/current/sql-createrole.html), so the following is written on the assumption that we're using the current Postgres term rather than looking for parity with the MySQL resource.

 - Implement a new Crossplane managed resource `Role` in `postgresql.sql.crossplane.io`.
 - Implement a new reconciler for `Role`.
 - Good initial support: all 'booleans' except `INHERIT` outlined in `CREATE ROLE` doco above, with at least plaintext password support.
 - `INHERIT` boolean only necessary with nested roles.

# Sketch

```yaml
kind: Role
metadata:
  name: example-role
spec:
  forProvider:

    # Either: Separate fields (Values are suggested defaults)
    superuser: false
    createDb: false
    createRole: false
    inherit: false
    login: true
    replication: false
    bypassRLS: false

    # OR: Implemented like the MySQL Grant type (I prefer this one)
    privileges:
      - NOSUPERUSER
      - NOCREATEDB
      - NOCREATEROLE
      - NOINHERIT
      - LOGIN
      - NOREPLICATION
      - NOBYPASSRLS

    # Subsequent functionality
    inRoleRefs:
      - name: other-role-1
        namespace: default
      - name: other-role-2
        namespace: default

    passwordSecretRef:
      name: example-pw
      namespace: default
      key: password

  writeConnectionSecretToRef:
    name: example-connection-secret
    namespace: default
```

Happy to look at a PR for this once API design is agreed upon.

e: `GRANT` also necessary of course, but would prefer to approach as 2 separate issues :)

benagricola:

This PR implements support for Postgres `ROLE` management, similarly to the existing MySQL `USER` support. A new resource type `Role` is added which allows configuration of role privileges, as well as setting a `ConnectionLimit` and plaintext password support via `PasswordSecretRef`.

Role membership (where roles can be members of other roles and inherit their granted permissions) is not configurable via this interface as it is more appropriately configured via `GRANT` management.

**NOTES** 

* This breaks from the way the MySQL `Grant` resources specify privileges as plaintext strings, to use booleans instead. This is because Postgres allows truthy, falsy and default values for each role privilege (i.e. `SUPERUSER`, `NOSUPERUSER` and _undefined_). Postgres defaulted values are lazy initialised back into the resource if unset. I don't think it's possible to properly represent this as a list of strings (and detect when updates are needed) because `rolsuper = f` from Postgres could mean both `NOSUPERUSER` defined by the user or _undefined_.
* There's scope to add additional tests to validate the SQL queries directly.

**EXAMPLE RESOURCE**

```yaml
kind: Role
metadata:
  name: example-role
spec:
  forProvider:
    connectionLimit: 10
    privileges:
      login: true
      createDb: true
      superUser: false
      ... # Rely on Postgres to populate the defaults for the other privileges.
  writeConnectionSecretToRef:
    name: example-role-secret
    namespace: default

```

Fixes #17

I have:

 - Appropriate tests have been copied from the MySQL `User` test suite.
 - Provider has been tested locally using `make dev` to confirm creation, update, and delete of new resources, as well as observation of existing resources. I used a local Postgres 13.1, but the code should work identically on all supported versions of Postgres (So 13 - 9.5).


negz:
Sorry I missed this one @benagricola! I'll aim to review it this week.

srueg (approved):
Looks good to me 👍

benagricola:
Closing as #21 already has these commits 👍

PR #22, opened 2021-02-02
Restrict mysql grant privilege strings to pattern

### LINKED ISSUE #18: MySQL Grant privileges arbitrary SQL injection
### What happened?
Looking through the MySQL `Grant` reconciler code, I noticed that `Spec.ForProvider.Privileges` is [passed directly](https://github.com/crossplane-contrib/provider-sql/blob/b4765ac1ca4b987b84c573c26f523e055b4729c9/pkg/controller/mysql/grant/reconciler.go#L257) into `createGrantQuery` and `fmt.Sprintf`'d into the SQL query. 

### How can we reproduce it?
Since the `Privileges` field is not sanity checked on input, it can contain any string the user chooses, example (untested):

```yaml
kind: Grant
metadata:
  name: inject-grant
spec:
  forProvider:
    privileges:
      - "ALL ON *.* TO '*'@'*'; RESTART; -- ",
```

Which would execute the following SQL queries, assuming relevant access: 

```
GRANT ALL ON *.* TO '*'@'*'; RESTART; --  (comment) ...
```

I don't believe `database/sql` does anything further to protect from this. 

Crossplane version: master
Provider version: master

### Fix
Privilege types cannot be quoted in the query, so should not be passed in directly from unsanitised user input.

I can think of a couple of options:
  * Sanitise the privileges list against a list of known privileges using `// +kubebuilder:validation:Enum=...` (This would ned to support multi-word enum values like `LOCK TABLES`...)
  * Switch input style somehow (boolean for each privilege?) so that privilege list is generated internally from user input

benagricola:
As discussed previously, the MySQL `Grant` implementation allowed arbitrary strings to be inserted into `privilege` queries. 

Although this was unlikely to expose significant SQL injection opportunities due to multiple statements being disabled in the mysql driver, it is safer to restrict the character set to only those used to represent privileges in MySQL. 

This commit filters the privileges at the API level using a custom string type, and adds role, database and privilege columns to status output for parity with the postgres `Grant` implementation in #21 (most of the code copied directly from that).

Fixes #18

I have:

Existing tests pass when called with the new custom string type, and local testing has been performed and confirmed against the previous behaviour using MariaDB 10.5.8.


srueg (approved):
Nice one, thanks for the contribution 👍

negz (approved):

negz:
Thanks @benagricola!

PR #26, opened 2021-02-23
Fix role privilege reconcile for non superusers

### LINKED ISSUE #25: provider-sql cannot update roles correctly on rds (or if it is not a superuser)
### What happened?

```
postgres=> CREATE ROLE "test" WITH NOSUPERUSER;
CREATE ROLE
postgres=> ALTER ROLE "test" WITH NOSUPERUSER;
ERROR:  must be superuser to alter superusers
```

On RDS, the primary user is not actually a superuser. It seems that if you include `NOSUPERUSER` in `ALTER ROLE`, then postgres throws an error. `provider-sql` does this because of the way clauses are managed. 

Clauses not set by the user are late-inited and then will be updated on the next reconciliation loop, containing all the clauses, even the negated clauses (`NO<...>`) that the role already does not posess.

### How can we reproduce it?
Create a `ProviderConfig` referring to a secret of a non-superuser that can create role and db's, and observe the error when creating a role.

Crossplane version: master
Provider-SQL version: master

### Proposed fix
We shouldn't blindly update all the clauses, including the late-inited ones, unless each one has changed. That way, the user will only get an error if they do not have permission to manage that particular clause (which should occur during the create phase anyway).

benagricola:
When a postgres role change is detected, `provider-sql` now compares the list of provided privileges against the observed list of privileges, and only updates the privileges which have changed. This avoids throwing errors for privileges which have not actually changed (and that our non-superuser `provider-sql` user cannot change after creation).

I've added tests to confirm:
1) Only updated privileges are put into an `ALTER ROLE ...` query
2) If no privileges are changed, no query is executed

Fixes #25

I have:

Additional tests have been written as per above, and the provider has been built and run locally against the RDS postgres instance that was exhibiting the original issue to confirm that the resource is reconciled correctly.

### DIFF
--- a/pkg/controller/postgresql/role/reconciler.go
+++ b/pkg/controller/postgresql/role/reconciler.go
@@ -19,7 +19,6 @@ package role
 import (
 	"context"
 	"fmt"
-	"sort"
 	"strings"
 	"time"
 
@@ -148,8 +147,23 @@ func privilegesToClauses(p v1alpha1.RolePrivilege) []string {
 	negateClause("REPLICATION", p.Replication, &pc)
 	negateClause("BYPASSRLS", p.BypassRls, &pc)
 
-	sort.Strings(pc)
 	return pc
+}
+
+func changedPrivs(existing []string, desired []string) []string {
+	out := []string{}
+
+	// The input slices here are outputted by privilegesToClauses above.
+	// Because these are created by repeated calls to negateClause in the
+	// same order, we can rely on each clause being in the same array
+	// position in the 'desired' and 'existing' inputs.
+
+	for i, v := range desired {
+		if v != existing[i] {
+			out = append(out, v)
+		}
+	}
+	return out
 }
 
 func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
This seems reasonable to me, but I don't have a ton of context on the implementation of this provider. @benagricola if you feel good about it feel free to move forward, otherwise @negz can take a look as well 👍

pkg/controller/postgresql/role/reconciler.go
```diff
@@ -148,10 +147,25 @@ func privilegesToClauses(p v1alpha1.RolePrivilege) []string {
...
+func changedPrivs(existing []string, desired []string) []string {
+	out := []string{}
+
+	// The input slices here are outputted by privilegesToClauses above.
+	// Because these are created by repeated calls to negateClause in the
+	// same order, we can rely on each clause being in the same array
+	// position in the 'desired' and 'existing' inputs.
+
+	for i, v := range desired {
+		if v != existing[i] {
```
negz, line 162: Could the slice of existing privs be shorter than the slice of desired privs, causing a panic here?

benagricola: Both lists are produced by `privilegesToClauses()` which will always create arrays of the same length and order. If a clause was added in a later version of the provider, then the list of privileges in `status.atProvider.PrivilegesAsClauses` would in theory be shorter.

I think that case is covered on upgrade because `Observe()` would always run first, so the correct length value should always be set, however - I suppose you could still edit the status field manually and cause a panic.

I'll add a length comparison just in case and error out 👍

negz (approved):

PR #33, opened 2021-03-10
Increase controller max concurrency

srueg:
On clusters with a high count of resources this significantly increases
the throughput of this provider.

Also see https://github.com/crossplane/crossplane/pull/2167

Increases the max concurrency of all controllers to 5.

I have:

Locally against a kind cluster.


srueg:
@negz Is this (increasing `MaxConcurrentReconciles`) something that you want to consider doing in `crossplane-runtime` to set a certain default value?

hasheddan (approved):
@srueg unfortunately since this is passed in at the controller level we can't enforce usage in `crossplane-runtime`. However, we could expose a constant that we recommend folks use. This is kind of like what we do with rate limiting (https://github.com/crossplane/crossplane-runtime/pull/243) where provider authors have to opt in. I am wondering if providing a higher-level wrapper around the controller manager would be useful in crossplane-runtime at this point if folks really just want to have the simplest experience of spinning up a provider 🤔 I do worry that folks would quickly outgrow any sort of abstraction that was that high level though.

benagricola (approved):
This looks good to me as the `maxConcurrency` setting is relatively modest. 

We open and close DB connections quite aggressively so I don't think the theoretically higher number of parallel connections will be an issue (at least not with only 6 managed resource types).

PR #36, opened 2021-04-09
Configure reconcile intervals

srueg:
Set the short- and long waits for all reconcilers to 10s and 10 minutes
respectively. This increases the responsiveness of newly created managed
resources and lowers the load on databases.

See also https://github.com/crossplane/crossplane/issues/2256

I have:

Locally against cluster.

### DIFF
--- a/pkg/controller/mysql/database/reconciler.go
+++ b/pkg/controller/mysql/database/reconciler.go
@@ -18,6 +18,7 @@ package database
 
 import (
 	"context"
+	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -61,6 +62,8 @@ func Setup(mgr ctrl.Manager, l logging.Logger) error {
 		resource.ManagedKind(v1alpha1.DatabaseGroupVersionKind),
 		managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), usage: t, newDB: mysql.New}),
 		managed.WithLogger(l.WithValues("controller", name)),
+		managed.WithShortWait(10*time.Second),
+		managed.WithLongWait(10*time.Minute),
 		managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))))
 
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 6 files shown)


negz:
I'm okay with this varying for this provider, but I wonder whether we should document it somewhere? I'm thinking that this one provider taking 10 minutes to notice and correct changes that are made out of band where all other providers take only one minute might be surprising.

benagricola:
I don't have an issue with setting the short and long waits, but I was unfamiliar with the `managed.With...` functions and in what cases the short and long wait is applied and... couldn't find them in controller runtime because they were [removed in 243](https://github.com/crossplane/crossplane-runtime/pull/243/files)!

Is it worth bumping the referenced crossplane-runtime version and using `WithPollInterval`and a rate limiter instead?

srueg:
Good point, we could then also use the already existing `--sync` flag to configure the poll interval. This can then be used via `ControllerConfig` and doesn't need any changes to the default.

srueg:
Looking at this again I don't really understand, why crossplane-runtime introduced this `PollInterval` instead of just using the manager's `SyncPeriod`?

```
// SyncPeriod determines the minimum frequency at which watched resources are
// reconciled. A lower period will correct entropy more quickly, but reduce
// responsiveness to change if there are many watched resources. Change this
// value only if you know what you are doing. Defaults to 10 hours if unset.
// there will a 10 percent jitter between the SyncPeriod of all controllers
// so that all controllers will not send list requests simultaneously.
SyncPeriod *time.Duration
```

Instead of returning `RequeueAfter` here: https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L789
It should be possible to just set `Requeue` and let `SyncPeriod` handle the timing?

/cc @hasheddan

negz:
> I don't really understand, why crossplane-runtime introduced this PollInterval instead of just using the manager's SyncPeriod?

I can't speak for @hasheddan  but I think it's entirely possible that we just forgot that was a thing when we introduced the `PollInterval`. I certainly forgot it was a thing when we were discussing how to build this.

negz:
@srueg It looks like there is in fact a (subtle) difference - let me respond over on https://github.com/crossplane/crossplane-runtime/issues/253.

srueg:
Thanks for the clarification, that makes sense to me 👍 

I created another PR to update crossplane-runtime and tune the poll `PollInterval` accordingly: https://github.com/crossplane-contrib/provider-sql/pull/37

PR #37, opened 2021-04-12
Update runtime and tune poll interval

srueg:
Update crossplane-runtime to latest version.
Set the poll interval for all reconcilers to 10 minutes.
This increases the responsiveness of newly created managed
resources and lowers the load on databases.

See also https://github.com/crossplane/crossplane/issues/2256

Replaces #36 

I have:

Locally against cluster.


srueg:
I tested this on our setup and it improves the responsiveness even further. Especially when deleting a resource, it won't wait for `shortWait` anymore, but reconcile immediately due to the new back-off behavior of crossplane-runtime 👍 

I think core Crossplane (especially the Composite reconciler) would also benefit a lot of this update.

benagricola (approved):
Pulled locally, tested and observed the expected backoff behaviour similar to [provider-aws/#544](https://github.com/crossplane/provider-aws/pull/544). 10 minute polling works well too.

It would be good in future if the poll interval were user-configurable as I think the correct value for this varies somewhat depending on use case, but that seems like it may be more of a concern of `controller-runtime` to implement across the board.

LGTM

srueg:
> It would be good in future if the poll interval were user-configurable as I think the correct value for this varies somewhat depending on use case, but that seems like it may be more of a concern of `controller-runtime` to implement across the board.

See https://github.com/crossplane/crossplane-runtime/issues/253

PR #35, opened 2021-04-01
Initial commit for extensions

linuxbsdfreak:
### Description of your changes

Fixes #

I have:

- [ ] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

This is a Draft PR. Testing is in progress

[contribution process]: https://git.io/fj2m9

PR #38, opened 2021-04-12
Allow database to be overridden in DSN from postgres constructor

### LINKED ISSUE #34: Allow custom db name to be passed to postgres client
Postgres has some namespaced resources that others have requested as a use-case, for example database extensions (#24).

Right now, all queries are executed against the `postgres` db as that is defaulted by the underlying driver. This can currently be controlled by setting the `PGDATABASE` environment variable via a `ControllerConfig` but this is provider-wide, and doesn't allow us to e.g. install an extension into a newly-created database.

We should allow the database to be set on an existing `postgresDB` instance (which would cause a modification of the DSN) so that a reconciler can choose to override the database name that queries are executed against. This will allow for further PR's adding support for Extensions, Grants against databases etc.

benagricola:
This is in preparation for further functionality where the connection database name for the PostgreSQL client needs to be controlled dynamically, e.g. where resources must be created under a particular database. We must _connect_ on a particular database to be able to create resources on that db (e.g. Extensions).

Fixes #34

I have:

There are no tests that cover the SQL client functionality directly. 

The existing test suite still passes, and I have tested this code on a local kind cluster, both with and without overriding the database. I've also confirmed the expected behaviour when using `PGDATABASE=blah`, and making sure this is still used if an empty string is set when calling `postgresql.New()`

### DIFF
--- a/pkg/clients/postgresql/postgresql.go
+++ b/pkg/clients/postgresql/postgresql.go
@@ -16,8 +16,10 @@ type postgresDB struct {
 	port     string
 }
 
-// New returns a new PostgreSQL database client.
-func New(creds map[string][]byte) xsql.DB {
+// New returns a new PostgreSQL database client. The default database name is
+// an empty string. The underlying pq library will default to either using the
+// value of PGDATABASE, or if unset, the hardcoded string 'postgres'.
+func New(creds map[string][]byte, database string) xsql.DB {
 	// TODO(negz): Support alternative connection secret formats?
 	endpoint := string(creds[xpv1.ResourceCredentialsSecretEndpointKey])
 	port := string(creds[xpv1.ResourceCredentialsSecretPortKey])
@@ -26,7 +28,8 @@ func New(creds map[string][]byte) xsql.DB {
 			string(creds[xpv1.ResourceCredentialsSecretUserKey]) + ":" +
 			string(creds[xpv1.ResourceCredentialsSecretPasswordKey]) + "@" +
 			endpoint + ":" +
-			port,
+			port + "/" +
+			database,
 		endpoint: endpoint,
 		port:     port,
 	}
--- a/pkg/controller/postgresql/database/reconciler.go
+++ b/pkg/controller/postgresql/database/reconciler.go
@@ -78,7 +78,7 @@ func Setup(mgr ctrl.Manager, l logging.Logger) error {
 type connector struct {
 	kube  client.Client
 	usage resource.Tracker
-	newDB func(creds map[string][]byte) xsql.DB
+	newDB func(creds map[string][]byte, database string) xsql.DB
 }
 
 func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
@@ -111,7 +111,7 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 		return nil, errors.Wrap(err, errGetSecret)
 	}
 
-	return &external{db: c.newDB(s.Data)}, nil
+	return &external{db: c.newDB(s.Data, "")}, nil
 }
 
 type external struct{ db xsql.DB }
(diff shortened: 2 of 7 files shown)


pkg/clients/postgresql/postgresql.go
```diff
@@ -3,38 +3,76 @@ package postgresql
...
 	}
+	db.formatDSN()
+	return &db
+}
+
+// formatDSN sets the DSN that this client will use
+// to connect to PostgreSQL.
+func (c *postgresDB) formatDSN() {
+	if c == nil {
+		panic(panicNilPtr)
```
srueg, line 50: Does this really add any benefit over just letting it panic with a nil dereference later on?

pkg/clients/postgresql/postgresql.go
```diff
@@ -3,38 +3,76 @@ package postgresql
...
+	}
+	c.database = db
+	c.formatDSN()
 }
 
 // ExecTx executes an array of queries, committing if all are successful and
 // rolling back immediately on failure.
-func (c postgresDB) ExecTx(ctx context.Context, ql []xsql.Query) error {
+func (c *postgresDB) ExecTx(ctx context.Context, ql []xsql.Query) error {
+	if c == nil {
```
srueg, line 73: As mentioned above; I don't think it's necessary to add these `c == nil` checks.

pkg/clients/postgresql/postgresql.go
```diff
@@ -3,38 +3,76 @@ package postgresql
...
 }
 
 // New returns a new PostgreSQL database client.
+// The default database name is an empty string.
+// The underlying pq library will default to either
+// using the value of PGDATABASE, or if unset, the
+// hardcoded db name 'postgres'.
+// The db name can be overridden later using SetDatabase,
+// which will trigger a re-format of the DSN.
 func New(creds map[string][]byte) xsql.DB {
```
srueg, line 34: Could we also just extend the `New()` with a `database` parameter?
IMHO this would make for a more ergonomic API than using a separate `SetDatabase()` call.

benagricola: We could allow the database to be set with `New()` as well, and that would be useful to add a field to `ProviderConfig` to set the default database name - which is effectively a nicer way to set `PGDATABASE` on the `ControllerConfig`. 

However - I don't think that helps with the core use-case I'm approaching here which is at runtime, we might want to resolve the database name to something after `Connect` has already been called.

It's worth thinking about this in combination with the `Extension` draft PR (#35). To install an `Extension` into a particular database on the instance, we have to connect to postgres on that database. The database itself might be created by crossplane as well, so might not exist when `Connect` is called if it is resolved as part of a Composition.

Since we don't actually connect to the instance until we're about to run a query, we can just override the DSN prior to executing it.

I thought doing this as a separate method would make sense, rather than modifying the existing query / exec / scan methods (which would've required changing their usage in the MySQL client as well).

I suppose we could allow the database name to be set in `New()` and then override the entire `xsql.DB` instance when it changes, which would avoid the additional `SetDatabase` method, but that would require us to store the credentials data somewhere outside `xsql.DB` to allow us to recreate the instance with the correct details.

pkg/clients/postgresql/postgresql.go
benagricola: Nope, force of habit - better to panic at the point of dereference as you said 👍

pkg/clients/postgresql/postgresql.go
srueg: What I don't understand is, we have all information we need in `Connect()` (and therefore when calling `New()`), no? Why would it be necessary to change the DB later on?
The `Connect()` function is called on every reconcile run and can use the information of the resource which is being reconciled (e.g. the `Extension`).

benagricola: Cripes... yeah you're right 🤦‍♂️ - not sure what I was thinking! 

We just `Connect()` with a DB name that sets the DSN, and if no 'override' DB name is set, we just continue on as normal. If we resolve a DB name, we know that value for the next time `Connect()` is called.

Back later with a fixed PR 😂

srueg (approved):

PR #39, opened 2021-04-16
PG extension create and drop with db parameter config

### LINKED ISSUE #24: Feature to install PG Extensions
Feature to install PG extensions as part of creating DB would be a nice
https://www.postgresql.org/docs/9.1/sql-createextension.html
https://www.highgo.ca/2019/07/22/how-to-add-extensions-to-pg-on-ubuntu/

This would help to cover the whole Post setup process after the DB instance is created. Including User and DB creation.

linuxbsdfreak:

Fixes #24 

I have:

Testing has been done by the example provided in the examples directory 


apis/postgresql/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Schema for extension install.
+	// +optional
+	Schema *string `json:"schema,omitempty"`
+
+	// Database for extension install.
+	// +optional
+	Database *string `json:"database,omitempty"`
+
+	// DatabaseRef references the database object this extension it for.
```
benagricola, line 48:
```suggestion
	// DatabaseRef references the database object this extension is for.
```

apis/postgresql/v1alpha1/extension_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// +optional
+	DatabaseRef *xpv1.Reference `json:"databaseRef,omitempty"`
+
+	// DatabaseSelector selects a reference to a Database this extension is for.
+	// +immutable
+	// +optional
+	DatabaseSelector *xpv1.Selector `json:"databaseSelector,omitempty"`
+}
+
+// A ExtensionSpec defines the desired state of a Extension.
```
benagricola, line 59:
```suggestion
// ExtensionSpec defines the desired state of an Extension.
```

Super nitpicky but "an Extension" reads cleaner to me than "a Extension", can we use that (or just "Extension" on its own)?

benagricola (commented):
Hey @linuxbsdfreak, thanks for putting the work in to understand the extension functionality in PostgreSQL and implementing it as a new type. 

Aside from the two comment nits above, I also noticed that there's a couple of untested paths in the reconciler `Connect` and `lateInit` methods, highlighted in yellow / red on [codecov](https://app.codecov.io/gh/crossplane-contrib/provider-sql/compare/39/tree/pkg/controller/postgresql/extension/reconciler.go).

Fixing these should just be a case of adding a test for each block, making sure that the code inside each if statement is executed as part of a test.

Other than that though, this is looking good and I'll try to give it a shot locally tomorrow to confirm the functionality works as expected 👍

pkg/controller/postgresql/extension/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+	if ref == nil {
+		return nil, errors.New(errNoSecretRef)
+	}
+
+	s := &corev1.Secret{}
+	if err := c.kube.Get(ctx, types.NamespacedName{Namespace: ref.Namespace, Name: ref.Name}, s); err != nil {
+		return nil, errors.Wrap(err, errGetSecret)
+	}
+
+	if cr.Spec.ForProvider.Database != nil {
```
benagricola, line 112: @linuxbsdfreak In addition to this, I think we need to return an error from `Connect` if `DatabaseRef` or `DatabaseSelector` are configured but `Database` is `nil`. If we don't do that, there's a chance that we will install an extension into the default DB rather  than the DB we want to, if one of the database references or selectors is not resolved instantly.

PR #40, opened 2021-04-19
Add sslmode option for PostgreSQL

cwesdorp:
Running PosgreSQL (locally on k8s) without SSL enabled is not
supported by lib/pq which has the default SSL mode on 'required'.

Added reading of sslmode from the ProviderConfig so that it's possible to divert from the default lib/pq configuration. 
The SSL mode is used to establish a connection to the managed resource and it is included in the connection secrets for created databases so that users also know changes to the sslmode.

I have:

There are no unit tests on parsing ProviderConfig secrets. Tested this running on a local cluster in combination with Postgres in its default config from the Bitnami helm chart and not passing sslmode in the ProviderConfig.

### DIFF
--- a/pkg/clients/postgresql/postgresql.go
+++ b/pkg/clients/postgresql/postgresql.go
@@ -3,7 +3,6 @@ package postgresql
 import (
 	"context"
 	"database/sql"
-	"fmt"
 
 	"github.com/crossplane-contrib/provider-sql/pkg/clients/xsql"
 	"github.com/lib/pq"
@@ -30,11 +29,11 @@ type postgresDB struct {
 // New returns a new PostgreSQL database client. The default database name is
 // an empty string. The underlying pq library will default to either using the
 // value of PGDATABASE, or if unset, the hardcoded string 'postgres'.
-func New(creds map[string][]byte, database string) xsql.DB {
+// The sslmode defines the mode used to set up the connection for the provider.
+func New(creds map[string][]byte, database string, sslmode string) xsql.DB {
 	// TODO(negz): Support alternative connection secret formats?
 	endpoint := string(creds[xpv1.ResourceCredentialsSecretEndpointKey])
 	port := string(creds[xpv1.ResourceCredentialsSecretPortKey])
-	sslmode := string(creds[sslModeKey])
 
 	dsn := "postgres://" +
 		string(creds[xpv1.ResourceCredentialsSecretUserKey]) + ":" +
@@ -44,7 +43,7 @@ func New(creds map[string][]byte, database string) xsql.DB {
 		database
 
 	if sslmode != "" {
-		dsn = fmt.Sprintf("%s?sslmode=%s", dsn, sslmode)
+		dsn = dsn + "?sslmode=" + sslmode
 	}
 
 	return postgresDB{
--- a/pkg/clients/postgresql/postgresql.go
+++ b/pkg/clients/postgresql/postgresql.go
@@ -3,6 +3,7 @@ package postgresql
 import (
 	"context"
 	"database/sql"
+	"fmt"
 
 	"github.com/crossplane-contrib/provider-sql/pkg/clients/xsql"
 	"github.com/lib/pq"
@@ -15,12 +16,15 @@ const (
 	// https://www.postgresql.org/docs/current/errcodes-appendix.html
 	// These are not available as part of the pq library.
 	pqInvalidCatalog = pq.ErrorCode("3D000")
+
+	sslModeKey = "sslmode"
 )
 
 type postgresDB struct {
 	dsn      string
 	endpoint string
 	port     string
+	sslmode  string
 }
 
 // New returns a new PostgreSQL database client. The default database name is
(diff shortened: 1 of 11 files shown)


srueg (approved):

pkg/clients/postgresql/postgresql.go
```diff
@@ -3,32 +3,47 @@ package postgresql
...
 }
 
 // New returns a new PostgreSQL database client.
 func New(creds map[string][]byte) xsql.DB {
 	// TODO(negz): Support alternative connection secret formats?
 	endpoint := string(creds[xpv1.ResourceCredentialsSecretEndpointKey])
 	port := string(creds[xpv1.ResourceCredentialsSecretPortKey])
+	sslmode := string(creds[sslModeKey])
+
+	dsn := fmt.Sprintf("host=%s port=%s user=%s password=%s",
```
benagricola, line 32: Switching to the preparsed space separated format avoids running `ParseURL` within `pq`, which contains some [escaping functionality](https://github.com/lib/pq/blob/b2901c7946b69f1e7226214f9760e31620499595/url.go#L46). 

My guess is that a password can contain special characters (quotes and backslashes according to the `pq` code) that could end up failing to parse without that escaping in place.

Would it be simpler to continue to use the old `dsn` format with a 'query string' argument (these are 'accrued' into the parsed string much like the other settings)? `?sslmode=...`

cwesdorp: I made the switch to a connection string because I passing query parameters didn't seem to work and [a comment](https://github.com/lib/pq/blob/b2901c7946b69f1e7226214f9760e31620499595/url.go#L16) for `ParseUrl` indicated the function to be there for backward compatibility. Since provider-sql wasn't supporting parameters it seemed save changing the format of the `dsn` string. 
I can retry using a connection URL, or is the `nurl.Parse(url)` already un-escaping special chars and is  `escaper.Replace()` putting them back?

cwesdorp: I have retested with the url format DSN and update the MR with this. Structure is code is equal, `sslmode` is only appended when provided.

benagricola:
@cwesdorp Is there a functional benefit from your perspective of this being within the connection secret vs being implemented as an additional field on `ProviderConfig`? 

There's no other resources that would output `sslmode` as a connection secret (yet, I guess) so unless the secret is created by hand, I assume this would be a static string outputted by a Composition, in a similar way to how the port might be hardcoded. 

As you need a `ProviderConfig` to reference a particular secret anyway, we have 2 places where `sslmode` might make sense.

I was unsure of this myself, so asked for input in slack ([thread here](https://crossplane.slack.com/archives/CEF5N8X08/p1618840086217700)), where @hasheddan put it rather succinctly:

> I generally think of the connection secret as "credentials to connect" and the provider config (or application that consumes a connection secret) as dictating "how you connect".

> Something like `sslmode` sounds like it could be "how" you connect though so that seems more reasonable to [put in] provider config

If this were to be moved into a field on `ProviderConfig` rather than within the secret, would it impact your usability in this particular case?

negz:
> Looking at resources.go it has keys for certificates, but no key to specify the SSL requirements.

FWIW those consts are more suggestions than requirements; you're welcome to define your own connection secret keys as appropriate. Though in this case I presume that would involve having to teach various SQL instance managed resources about the new SSL mode key.

I've only skimmed this thread, but my two cents would be that it likely makes the most sense to keep the SSL mode in the `ProviderConfig`. One justification for adding it to the provider credentials secret would be if it _were_ being automatically derived from an SQL instance (e.g. an RDS instance MR was automatically figuring out the SSL mode needed to connect to it and writing that to its connection secret). Though even then you could probably use an XR to communicate that via the `ProviderConfig`; e.g. have your Composition create an SSL enabled RDS instance and also a `ProviderConfig` with SSL enabled that points at it.

cwesdorp:
@negz okay, clear. Following this reasoning, would the SSL mode is also not part of the Secret written by the Role based on the [GetConnectionDetails](https://github.com/crossplane-contrib/provider-sql/blob/4dc73373342863176cb7d6ff04c075b350455453/pkg/clients/postgresql/postgresql.go#L124)?

benagricola:
> @negz okay, clear. Following this reasoning, would the SSL mode is also not part of the Secret written by the Role based on the [GetConnectionDetails](https://github.com/crossplane-contrib/provider-sql/blob/4dc73373342863176cb7d6ff04c075b350455453/pkg/clients/postgresql/postgresql.go#L124)?

There's 2 parts of this I think - if you're using newly generated connection details with _crossplane_ (e.g. you generated a new restricted user and want to now have crossplane manage something as that particular user) then you need a separate `ProviderConfig` anyway to tell crossplane to use the new secret - so you can set the `SSLMODE` there.

In cases where the secret is used directly by an application, then whether a server needs SSL enabled or not is something that your Database composition should be able to work out, and you could return it as your own custom key from the composition using [`connectionDetails`](https://crossplane.io/docs/v1.1/concepts/composition.html) (search for that exact string on the linked page for an example).

I think that way it would only be necessary to implement it on `ProviderConfig` where you want to be able to use crossplane to manage resources across instances or databases which may be both SSL enabled and disabled.

cwesdorp:
@benagricola The SSLMode attribute has moved to ProviderConfig. It is now passed as a parameter, just as the database name which also has become a parameter. The branch is updated to the latest state of master as well. Please let me know what you think.

pkg/clients/postgresql/postgresql.go
```diff
@@ -116,6 +128,7 @@ func (c postgresDB) GetConnectionDetails(username, password string) managed.Conn
 		xpv1.ResourceCredentialsSecretPasswordKey: []byte(password),
 		xpv1.ResourceCredentialsSecretEndpointKey: []byte(c.endpoint),
 		xpv1.ResourceCredentialsSecretPortKey:     []byte(c.port),
+		sslModeKey:                                []byte(c.sslmode),
```
turkenh, line 131: Looks like we ended up not reading from connection details in favor of provider config spec. This like should be removed.

Duologic:
@cwesdorp What is the reason for closing this PR?

PR #62, opened 2022-01-23
Escape userinfo in PostgreSQL DSN using net/url UserInfo

### LINKED ISSUE #61: PostgreSQL client fails to connect if password contains characters that need to be URL encoded
### What happened?

My (auto-generated) postgres password contained an '^' character. As a result, provider-sql was not able to connect to the database. The following error was present in the logs (password, host and dbname redacted):

Warning    {"object": {"kind":"Extension","name":"isvanilla-uuid-extension","uid":"9c74ceb9-5
f04-4b4c-b508-00a820cea033","apiVersion":"postgresql.sql.crossplane.io/v1alpha1","resourceVersion":"300047"}, "reason": "CannotObserveExternalResource", "message": "ca
nnot select extension: parse \"postgres://postgres:[password]@[host]:5432/[dbname]\": net/ur
l: invalid userinfo"}

Looking at the code that produces the DSN URL, I think there is a need to escape the string to ensure the username and password are safely encoded for use in a URL. 

```
func New(creds map[string][]byte, database string) xsql.DB {
	// TODO(negz): Support alternative connection secret formats?
	endpoint := string(creds[xpv1.ResourceCredentialsSecretEndpointKey])
	port := string(creds[xpv1.ResourceCredentialsSecretPortKey])
	return postgresDB{
		dsn: "postgres://" +
			string(creds[xpv1.ResourceCredentialsSecretUserKey]) + ":" +
			string(creds[xpv1.ResourceCredentialsSecretPasswordKey]) + "@" +
			endpoint + ":" +
			port + "/" +
			database,
		endpoint: endpoint,
		port:     port,
	}
}
```

### How can we reproduce it?

Attempt to connect to a Postgres database using a password containing a '^' character

Crossplane version: 1.6.1
Running in EKS on AWS (K8s 1.21)

jdotw:

Fixes #61 

If a password contains a character that needs to be percent-encoded for use in the DSN, the connection to postgres will fail with a `net/url: invalid userinfo` error. 

This change first refactors the creation of the DSN string to allow unit testing, then uses net/url's UserInfo to encode the username:password portion of the DSN. This will apply percent-encoding appropriately. 

Included a unit test that checks for proper encoding of password with a '^' character which is valid in PostgreSQL, but causes provider-sql to fail to connect without this fix. 

I have:

Added unit test to confirm string encoding

Ran 'make reviewable test'


Duologic (approved):
I'm running into the same issue, I have an `@` in the password. The net/url.UserPassword function is also what postgres_exporter uses. 

LGTM

negz:
Closed and re-opened in the hopes of triggering CI.

pkg/clients/postgresql/postgresql.go
```diff
@@ -30,18 +31,28 @@ func New(creds map[string][]byte, database string) xsql.DB {
...
-			port + "/" +
-			database,
+		dsn:      DSN(username, password, endpoint, port, database),
 		endpoint: endpoint,
 		port:     port,
 	}
 }
 
+// DSN returns the DSN URL
+func DSN(username string, password string, endpoint string, port string, database string) string {
```
negz, line 44:
```suggestion
func DSN(username, password, endpoint, port, database string) string {
```

Just a nitpick. :)

negz (approved):

negz:
Whoops - I may have approved too soon. Looks like a unit test is failing.

Also, would we need a similar change for MySQL?

jdotw:
> Whoops - I may have approved too soon. Looks like a unit test is failing.
> 
> Also, would we need a similar change for MySQL?

Likely.. I've opened #70 to track this.

PR #63, opened 2022-02-04
feat(mysql): add resource options

Duologic:
I'd like to extend the mysql part of the provider to support resource options when
creating users. (https://dev.mysql.com/doc/refman/8.0/en/user-resources.html). In
particular I have a usecase for `MAX_USER_CONNECTIONS`.

I've used the 'privileges' concept from the postgresql part as an example to write this
feature.

I have:

- [x] Added a unit test to verify the idempotency.
- [x] Added parameters to the example and ran `make e2e` successfully to test the resource model and queries.


pkg/controller/mysql/user/reconciler.go
```diff
@@ -222,14 +222,13 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 
-	cr.SetConditions(xpv1.Available())
-
 	cr.Status.AtProvider.ResourceOptionsAsClauses = resourceOptionsToClauses(observed.ResourceOptions)
 
+	cr.SetConditions(xpv1.Available())
+
 	return managed.ExternalObservation{
-		ResourceExists:          true,
-		ResourceLateInitialized: lateInit(observed, &cr.Spec.ForProvider),
```
Duologic: I mimicked the lateInit from the postgres role reconciler, but apparently I don't understand the implications of this. It caused the User object to never become ready in the integration test. I've removed this in the PR here for now, not sure if needed.

negz (commented):
This looks good to me, but the functionality it's orchestrating is well past my level of MySQL knowledge.

apis/mysql/v1alpha1/user_types.go
```diff
@@ -40,10 +40,36 @@ type UserParameters struct {
 	// for this user. If no reference is given, a password will be auto-generated.
 	// +optional
 	PasswordSecretRef *xpv1.SecretKeySelector `json:"passwordSecretRef,omitempty"`
+
+	// ResourceOptions sets account specific resource limits.
+	// See https://dev.mysql.com/doc/refman/8.0/en/user-resources.html
+	// +optional
+	ResourceOptions ResourceOptions `json:"resourceOptions,omitempty"`
```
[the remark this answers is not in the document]
Duologic: Done.

negz (approved):
LGTM, though I would appreciate a review from someone (@jdotw?) who is more familiar with the subject matter.

package/crds/mysql.sql.crossplane.io_users.yaml
```diff
@@ -70,6 +70,22 @@ spec:
...
+                      maxQueriesPerHour:
+                        description: MaxQueriesPerHour sets the number of queries an account can issue per hour
+                        type: integer
+                      maxUpdatesPerHour:
+                        description: MaxUpdatesPerHour sets the number of updates an account can issue per hour
+                        type: integer
+                      maxUserConnections:
+                        description: MaxUserConnections sets The number of simultaneous connections to the server by an account
+                        type: integer
+                    type: object
```
jdotw, line 88: Not to sound like a broken record... but should there be defaults here to avoid new comers from being stuck wondering what values are "ok"?

jdotw: Optional but you are essentially enforcing a default in the actual code of the provider, right? Or did I read that wrong?

Duologic: Can you point out where that is? I'm probably overlooking something.

jdotw: Sorry man, I failed to read for detail. I remembered seeing some specific values when I scanned the diff but now as I look back I see they're from the example yaml. *face palm*

PR #67, opened 2022-03-04
Update build tools to latest version

Duologic:
### Description of your changes

The build tools are a tad bit outdated, with this PR I'm looking at the Golang toolchain.

In follow ups I'd like to update the tools used by `make e2e` and the libraries in
`go.mod`.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

`make reviewable test`

`make e2e`

[contribution process]: https://git.io/fj2m9

PR #69, opened 2022-03-14
feat(postgresql): Add role configuration parameters

### LINKED ISSUE #68: role.postgresql: Support for configuration parameters
Need to be able to set role configuration parameters. This is useful for instance together with the pgaudit extension where you can do `alter role myrole set pgaudit.log='All'` to audit the actions of myrole.

https://www.postgresql.org/docs/current/sql-alterrole.html

I suggest introducing configurationParameters where name value pairs can be defined and applied to the role.
```
kind: Role
metadata:
  name: parent-role
spec:
  forProvider:
    connectionLimit: 10
    privileges:
      login: true
    configurationParameters:
      - name: 'pgaudit.log'
        value: 'All'
      - name: 'search_path'
        value: '"$user",public'
```

petteja:
This adds role configuration parameters to the PostgreSQL Role CRD. This allows to set configuration parameters like `pgaudit.log='All'` or `search_path="$user, public"` to the role.

Fixes #68

I have:

This has been tested manually in my test environments by creating users using crossplane, and also by taking ownership over current users using `crossplane.io/external-name` annotations.

I've also added a couple of go tests for this.


apis/postgresql/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
Duologic, line 1: The CI error is probably because of this line, the go version needs updating, I've got a PR for this here: https://github.com/crossplane-contrib/provider-sql/pull/67/files

petteja: Yes, most likely. Should I remove the line, or should we wait for that PR to get merged?

jdotw: #67 has been merged, but it looks like the test is still failing. @petteja can you take a look?

Duologic: This branch now just needs a rebase to pull the new github actions (and probably resolve merge conflicts from other changes).

petteja: Merged in changes from master

petteja: Ran `make reviewable test e2e` on a different PC and it worked just fine. :man_shrugging:

apis/postgresql/v1alpha1/role_types.go
```diff
@@ -83,13 +83,28 @@ type RoleParameters struct {
 	// for this role. If no reference is given, a password will be auto-generated.
 	// +optional
 	PasswordSecretRef *xpv1.SecretKeySelector `json:"passwordSecretRef,omitempty"`
+
+	// ConfigurationParameters to be applied to the role. If specified, any other configuration parameters set on the
+	// role in the database will be reset.
+	//
+	// See https://www.postgresql.org/docs/current/runtime-config-client.html for some available configuration parameters.
+	// +optional
+	ConfigurationParameters []RoleConfigurationParameter `json:"configurationParameters,omitempty"`
```
Duologic, line 92: Nit: Kubernetes API guidelines require nil-able optional fields.

Perhaps a `RoleConfigurationParameterList` might be useful here:

```
type RoleConfigurationParameterList []RoleConfigurationParameter
```

Then we can do something like this:

```
ConfigurationParameters *RoleConfigurationParameterList `json:"configurationParameters,omitempty"`
```

Duologic (approved):
One nit, I learned about that myself recently, other than that the PR looks good.

apis/postgresql/v1alpha1/role_types.go
petteja: Cool, I wasn't aware of this. However I can't get `make generate` to work if I make that change. Not too familiar with this stuff :D 
`angryjet: error: error loading packages using pattern ./...: git/petteja/provider-sql/apis/postgresql/v1alpha1/zz_generated.deepcopy.go:766:10: cannot use new([]RoleConfigurationParameter) (value of type *[]RoleConfigurationParameter) as *RoleConfigurationParameterList value in assignment`

Any clues?

petteja: Couldn't get it working with type aliasing, might be a bug in deepcopy generation. Changed to pointer type without aliasing.

Duologic: Gotcha, great that you got it working. Will review again later today

petteja: The test seems to fail because of https://github.com/crossplane-contrib/provider-sql/pull/73

Duologic (approved):

PR #80, opened 2022-04-06
Mysql default port

### LINKED ISSUE #76: Use default Mysql port if not specified in the connection secret

If a secret containing connection parameters for Mysql server doesn't contain `port`, the connection will timeout when attempting to use the provider, e.g. when creating a user:

```
CannotObserveExternalResource  user/test-user  cannot select user: dial tcp 10.x.x.x:0: i/o timeout
```

The problem is that provider-aws doesn't write `port` data for DB cluster (maybe for others as well). While it makes sense to address this in provider-aws, I also believe it makes sense to set a default port in provider-sql, as user's intention would likely be to connect over 3306 (instead of 0).

Example secret created by provider-aws:

```
apiVersion: v1
data:
  endpoint: [...]
  password: [...]
  username: [...]
kind: Secret
metadata:
  creationTimestamp: "2022-03-22T11:28:56Z"
  name: test-db-connection
  namespace: default
  ownerReferences:
  - apiVersion: rds.aws.crossplane.io/v1alpha1
    controller: true
    kind: DBCluster
    name: cluster-kw87c
    uid: [...]
  resourceVersion: [...]
  uid: [...]
type: connection.crossplane.io/v1alpha1
```

Assume that the absence of port in the connection secret means that the connection attempt should happen over the default port.

I'd be happy to submit a PR if this proposal makes sense. As far as I can tell the required change is in `pkg/clients/mysql/mysql.go`, where we could put a default port constant, and use it if the port is an empty string.

sendmars:

When connection secret doesn't contain `port`, this change assumes default port for MySQL, 3306.

This is needed because provider-aws doesn't write port (by default) to the connection secret for Aurora DB cluster. Provider-sql shouldn't accept the empty string as it is an invalid value for port, and assume that a missing value should default to the standard port.

Fixes #76

I have:

[Testing incomplete]
I ran `make run` locally, which caused a local reconciler to run in parallel with the one in the kubernetes cluster.

When describing the `user.mysql` on the kubernetes cluster, I see (messages only, for readability):

```
cannot select user: dial tcp 10.0.x.x:3306: i/o timeout         # local, patched reconciler
cannot select user: dial tcp 10.0.x.x:0: i/o timeout            # reconciler in the cluster
```

The change sets the port correctly. The connection to db_host:3306 is open, so I would expect that the user is created. The problem seem not to be related to my change, but I would need some assistance in figuring this out. The debug output contains (newlines added manually):

```
2022-04-05T23:46:44.471+0100    DEBUG   provider-sql    Cannot observe external resource
{
"controller": "managed/user.mysql.sql.crossplane.io",
"request": "/cluster-test-user",
"uid": "...",
"version": "...",
"external-name": "cluster-test-user",
"error": "cannot select user: dial tcp 10.0.x.x:3306: i/o timeout",
"errorVerbose": "dial tcp 10.0.x.2x:3306: i/o timeout
      cannot select user
      github.com/crossplane-contrib/provider-sql/pkg/controller/mysql/user.(*external).Observe
          /workdir/opensource/provider-sql/pkg/controller/mysql/user/reconciler.go:216 ..."
```

I checked that the port 3306 is part of `c.db` in `Observe`. I suspect that the issue with connection comes from the interaction of the two reconcilers.

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -39,7 +39,7 @@ func main() {
 		app            = kingpin.New(filepath.Base(os.Args[0]), "SQL support for Crossplane.").DefaultEnvars()
 		debug          = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
 		syncPeriod     = app.Flag("sync", "Controller manager sync period such as 300ms, 1.5h, or 2h45m").Short('s').Default("1h").Duration()
-		leaderElection = app.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
+		leaderElection = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
 	)
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
--- a/pkg/clients/mysql/mysql.go
+++ b/pkg/clients/mysql/mysql.go
@@ -14,7 +14,8 @@ import (
 )
 
 const (
-	errNotSupported = "%s not supported by mysql client"
+	errNotSupported  = "%s not supported by mysql client"
+	defaultMySQLPort = "3306"
 )
 
 type mySQLDB struct {
@@ -28,6 +29,9 @@ func New(creds map[string][]byte) xsql.DB {
 	// TODO(negz): Support alternative connection secret formats?
 	endpoint := string(creds[xpv1.ResourceCredentialsSecretEndpointKey])
 	port := string(creds[xpv1.ResourceCredentialsSecretPortKey])
+	if port == "" {
+		port = defaultMySQLPort
+	}
 	return mySQLDB{
 		dsn: fmt.Sprintf("%s:%s@tcp(%s:%s)/",
 			creds[xpv1.ResourceCredentialsSecretUserKey],


Duologic (commented):
This could work as a stop-gap but has a very limited usecase.

I've touched a similar problem where I wanted to set the endpoint to local pod (cloudsql-proxy on GCP), I managed to workaround it by passing in the endpoint/port through a label on a Composition.

However it can be a bit more generic, perhaps we can extend the ProviderConfigs with overrides for host/port? This can be fairly generic for all 3 database types. WDYT?

luisdavim:
> This could work as a stop-gap but has a very limited usecase.
> 
> I've touched a similar problem where I wanted to set the endpoint to local pod (cloudsql-proxy on GCP), I managed to workaround it by passing in the endpoint/port through a label on a Composition.
> 
> However it can be a bit more generic, perhaps we can extend the ProviderConfigs with overrides for host/port? This can be fairly generic for all 3 database types. WDYT?

I think that extending the `ProviderConfigs` is a good idea but I don't see any issue with the change being proposed here, falling back to the default port when no port is provided makes sense to me.

Duologic:
I'd be up for merging this if someone can rebase.

pierluigilenoci:
@sendmars please rebase.

chlunde:
@sendmars could you rebase this PR? Thanks!

Otherwise I'll close it to clean up the PR backlog, but anyone interrested could recreate a similar PR. It would be nice to ensure all supported databases behave the same way, at least if that makes sense for all.

chlunde:
Closing to clean up PR backlog as this isn't mergeable as-is. Feel free to recreate the PR

PR #82, opened 2022-04-07
initial pass at adding tls support to mysql

### LINKED ISSUE #70: Mysql client may fail to connect if password contains characters that need to be URL encoded
Following on from #61 , the same issue is likely impacting MySQL.

Both use a DSN for the connection string, but there isn't URI-encoding of the user-specified password in MySQL such as was added for PostgreSQL in #62

### LINKED ISSUE #81: Add TLS support to MySQL Connections

I would like to enable TLS on MySQL connections.

The `tls` value is supported in the [go-sql-driver/mysql](https://github.com/go-sql-driver/mysql#tls) package, it just needs to be added as an option to this provider. I am working on a PR for this.

MattMencel:

Enables `tls` for MySQL connections. Sets `preferred` as the default which will fall back to non-tls if unsupported by the MySQL instance.

"Fixes #70":
"Fixes #81":

I have:

`make reviewable`


pkg/clients/mysql/mysql.go
```diff
@@ -21,24 +21,36 @@ type mySQLDB struct {
...
-			port),
+		dsn:      dsn,
 		endpoint: endpoint,
 		port:     port,
+		tls:      tls,
 	}
 }
 
+// DSN returns the DSN URL
+func DSN(username, password, endpoint, port, tls string) string {
```
Duologic, line 45: As you are touching this, would you consider improving it a little bit? With PostgreSQL we noticed the credentials weren't properly escaped, this was fixed in https://github.com/crossplane-contrib/provider-sql/pull/62 and should probably also be implemented for MySQL (issue https://github.com/crossplane-contrib/provider-sql/issues/70)

No prob if you don't want to, we can pick it up in another PR.

Duologic (commented):
The tests seem to fail, can you have a look? Note that the golangci-lint version might differ: https://github.com/crossplane-contrib/provider-sql/blob/2f501aa3a01ea869c3257731236c992423d95827/.github/workflows/ci.yml#L13-L15

pkg/clients/mysql/mysql.go
MattMencel: I think I have this updated for the escaping.

MattMencel:
> The tests seem to fail, can you have a look? Note that the golangci-lint version might differ:
> 
> https://github.com/crossplane-contrib/provider-sql/blob/2f501aa3a01ea869c3257731236c992423d95827/.github/workflows/ci.yml#L13-L15

I'm trying to get these tests to run in my fork's Github Actions so I can debug further. Same as Alex in #84 mentions, I'm not sure what to do to test locally. Maybe a quick guide on how to run them locally?

Duologic:
`make reviewable` fails when I check out your branch, following make targets should changes files that fix the CI:

```
make fmt.simplify
make generate
```

MattMencel:
Both those commands result in this...

```
> make fmt.simplify
make: *** No rule to make target `fmt.simplify'.  Stop.
```

MattMencel:
I bumped `golangci-lint` to 1.45 in the Github Action. It doesn't seem like there's a simple way to downgrade it on a Mac with Homebrew. All the tests are passing in my fork now.

Also, the CI pipeline has golang 1.17 pinned, but the `go.mod` file is still at 1.13?

Duologic:
I had a look at bumping the golangci-lint before but it is also set in the build tools. I think it is best to keep those aligned. https://github.com/upbound/build/blob/master/makelib/golang.mk#L99

If we want to bump versions, I'd suggest doing that in a separate PR so we can also adress the versions in the upstream repo.

MattMencel:
@Duologic I reverted the `golangci-lint` change

Duologic:
I'm attempting to fix the lint CI job in https://github.com/crossplane-contrib/provider-sql/pull/88

Duologic:
#88 is merged, do you mind rebasing?

MattMencel:
@Duologic all the tests passed in my fork

apis/mysql/v1alpha1/provider_types.go
```diff
@@ -26,6 +26,14 @@ import (
...
 	// Credentials required to authenticate to this provider.
 	Credentials ProviderCredentials `json:"credentials"`
+	// tls=true enables TLS / SSL encrypted connection to the server.
+	// Use skip-verify if you want to use a self-signed or invalid certificate (server side)
+	// or use preferred to use TLS only when advertised by the server. This is similar
+	// to skip-verify, but additionally allows a fallback to a connection which is
+	// not encrypted. Neither skip-verify nor preferred add any reliable security.
+	// +kubebuilder:validation:Enum=true;skip-verify;preferred
+	// +optional
+	TLS string `json:"tls"`
```
Duologic, line 36: Nit: This should be *string. Kubernetes API guidelines require nil-able optional fields.

MattMencel: Fixed

Duologic:
I think we need to revert the change on the build submodule and follow master.

MattMencel:
> I think we need to revert the change on the build submodule and follow master.

Yeah I tried to debug it yesterday but I wasn't getting anywhere. So change `*string` back to `string` you're saying?

MattMencel:
> @MattMencel I managed to get it working in #91, you can cherry-pick my commits if you like.

@Duologic cherry-picked

Duologic (approved):

PR #83, opened 2022-04-08
Support mysql-grants by table

atarax:

Introduces support for grants by individual tables in mysql

I have:

Code was tested in local-setup with kind-cluster created with `make dev`. A local mysql-instance was connected to the provider and manual verification has been done. 
Unit-tests were part of standard-procedure.
I did *not* adjust e2e. I ran these but i think i'm missing something because it's failing with ```Error from server (BadRequest): container "provider-sql" in pod "provider-sql-provider-sql-ffb4544-6kv29" is waiting to start: ErrImageNeverPull```

### DIFF
--- a/pkg/controller/mysql/grant/reconciler.go
+++ b/pkg/controller/mysql/grant/reconciler.go
@@ -216,23 +216,8 @@ func (c *external) getPrivileges(ctx context.Context, username, dbname string, t
 		}
 		p := parseGrant(grant, dbname, table)
 
-		/*
-			if p != nil && privileges != nil {
-				// Found more than one grant for this user/DB pair.
-				// This probably shouldn't happen because MySQL groups privileges
-				// for the same user/DB pair in a single grant.
-				// In any case we want to update and ensure the privileges of our grant.
-				return nil, &managed.ExternalObservation{
-					ResourceExists:   true,
-					ResourceUpToDate: false,
-				}, nil
-			}
-			privileges = p
-		*/
-
-		// NOTE: this breaks the previous "safety"-switch for the same user/DB pair in a single grant
-		// but when managing tables in grants we need to break as soon as we found something
 		if p != nil {
+			// found the grant we were looking for
 			privileges = p
 			break
 		}
--- a/pkg/controller/mysql/grant/reconciler.go
+++ b/pkg/controller/mysql/grant/reconciler.go
@@ -61,7 +61,7 @@ const (
 )
 
 var (
-	grantRegex = regexp.MustCompile("^GRANT (.+) ON `(.+)`\\.\\* TO .+")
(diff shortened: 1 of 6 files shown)


pkg/controller/mysql/grant/reconciler.go
```diff
@@ -206,19 +214,30 @@ func (c *external) getPrivileges(ctx context.Context, username, dbname string) (
...
-			// This probably shouldn't happen because MySQL groups privileges
-			// for the same user/DB pair in a single grant.
-			// In any case we want to update and ensure the privileges of our grant.
-			return nil, &managed.ExternalObservation{
-				ResourceExists:   true,
-				ResourceUpToDate: false,
-			}, nil
+		p := parseGrant(grant, dbname, table)
+
+		/*
```
atarax, line 219: In my opinion this block is not needed at all. Suppose there are two rows found, even when we send an `ResourceUpToDate: false` there is still absolutely no guarantee that we will end up in desired state. 
At least i fail to see the value added and unfortunately this defensive logic makes it impossible here to use tables. If we use tables we will always have multiple rows in the result (don't see a way to filter them upfront) and should pick the first combination that matches as there should be no other matches of interest.
Any thoughts? :)

Duologic: Your reasoning makes sense, feel free to remove it.

Duologic (commented):
LGTM

Note the failing CI lint job, we use  golangci-lint v1.31 as we follow upstream build/repository, this might explain the failure.

Duologic:
I'm attempting to fix the lint CI job in https://github.com/crossplane-contrib/provider-sql/pull/88

pkg/controller/mysql/grant/reconciler.go
tkaesserfm: removed

atarax: removed

atarax:
Thanks for your suggestion regarding e2e-tests. It's working fine now: ```11:30:12 [ OK ] integration tests passed```

Duologic:
I'm also fixing the e2e in #88, I noticed it was still running against Crossplane v0.14.0 :hand_over_mouth:

atarax:
@Duologic omw :)

Duologic (approved):

alexbowers:
@Duologic Any chance we can get a tag of the latest release with this in it?

tkaesserfm:
@alexbowers `crossplane/provider-sql:v0.4.1-17.g3fcd1c9` contains it :)

dudicoco:
@atarax any plans on adding this for postgres as well?

tkaesserfm:
@dudicoco not currently :/

PR #86, opened 2022-04-11
Set the default PostgreSQL sslMode to `verify-full`

### LINKED ISSUE #85: ProviderConfig can be deleted before ProviderConfigUsage is getting cleaned
### What happened?

The ProviderConfig of `postgresql.sql.crossplane.io` is getting deleted *before* associated ProviderConfigUsage is getting removed.

It leads to a Database resource that is using this ProviderConfig being stuck in an orphaned state.

### How can we reproduce it?

Create ProviderConfig like

```
kind: ProviderConfig
metadata:
  name: smalpe-db-bd848
spec:
  credentials:
    connectionSecretRef:
      name: azpgserver-conn-str
      namespace: test
    source: PostgreSQLConnectionSecret
  defaultDatabase: postgres
```

Create Database referencing this ProviderConfig
```
kind: Database
metadata:
  name: smalpe-db-bd848
spec:
  forProvider:
    allowConnections: true
    connectionLimit: -1
    encoding: UTF8
    isTemplate: false
    lcCType: English_United States.1252
    lcCollate: English_United States.1252
    owner: masteradmin
    tablespace: pg_default
  providerConfigRef:
    name: smalpe-db-bd848
```

Observe that associated ProviderConfigUsage was created

```
k get providerconfigusages.postgresql.sql.crossplane.io
NAME                                   AGE     CONFIG-NAME       RESOURCE-KIND   RESOURCE-NAME
1366b088-bd57-4286-864a-de44ef90f599   26m     smalpe-db-bd848   Database        smalpe-db-bd848
```

Delete ProviderConfig
```
k delete providerconfigs.postgresql.sql.crossplane.io smalpe-db-bd848
providerconfig.postgresql.sql.crossplane.io "smalpe-db-bd848" deleted
```

Observer that ProviderConfigUsage is still around
```
k get providerconfigusages.postgresql.sql.crossplane.io
NAME                                   AGE     CONFIG-NAME       RESOURCE-KIND   RESOURCE-NAME
1366b088-bd57-4286-864a-de44ef90f599   27m     smalpe-db-bd848   Database        smalpe-db-bd848
```

Try to delete the Database resource
```
k delete database.postgresql.sql.crossplane.io/smalpe-db-bd848
```

Observer that Database deletion is stuck with

```
k describe database.postgresql.sql.crossplane.io/smalpe-db-bd848
....
  Warning  CannotConnectToProvider      12s (x12 over 33s)  managed/database.postgresql.sql.crossplane.io  cannot get ProviderConfig: ProviderConfig.postgresql.sql.crossplane.io "smalpe-db-bd848" not found
``` 

This issue heavily affects the situation when ProviderConfig and Database are a part of the same Composition.

When the associated Claim is deleted the ProviderConfig is getting deleted first and the Database deletion is stuck.

Crossplane version:  1.6.3-up.1

ulucinar:

Fixes #85

This PR proposes to set the default PostgreSQL sslMode to `verify-full`. We can also consider having `spec.sslMode` of `ProviderConfig.postgresql`  as really optional and leave the defaults to the driver. For instance, [libpq defaults to `prefer`](https://www.postgresql.org/docs/current/libpq-ssl.html), which is not recommended for security reasons. However, for improved security, this PR proposes `verify-full` as the default. 

I have:


apis/postgresql/v1alpha1/provider_types.go
```diff
@@ -33,7 +33,8 @@ type ProviderConfigSpec struct {
 	// Defines the SSL mode used to set up a connection to the provided
 	// PostgreSQL instance
 	// +kubebuilder:validation:Enum=disable;require;verify-ca;verify-full
-	// +optional
+	// +kubebuilder:default=verify-full
+	// +kubebuilder:validation:Required
 	SSLMode string `json:"sslMode"`
```
turkenh, line 38: I believe we should set it as optional (with omit-empty) since we have a default, e.g. 

https://github.com/crossplane/crossplane/blob/7d01448f8a52e0f3a13add4fc0f81028fcd13400/apis/pkg/v1/package_types.go#L31

ulucinar: Thanks @turkenh, converted it to an optional field.

turkenh (approved):

Duologic:
I'd argue to follow the upstream default as it is more predictable, is there a change request upstream to change the default perhaps?

ulucinar:
> I'd argue to follow the upstream default as it is more predictable, is there a change request upstream to change the default perhaps?

Hi @Duologic,
I have not seen anything upstream but here is a documentation for the ssl mode:
https://www.postgresql.org/docs/current/libpq-ssl.html

Please also see the note in that document reproduced here for convenience:
```
... Relying on this behavior is discouraged, and applications that need certificate validation should always use verify-ca or verify-full.
```

As mentioned in the description of the PR, while I agree aligning with the downstream default would be less controversial but I preferred the _secure by default_ approach and chose the strictest policy. Also with many modern TLS client libraries, what's described for `verify-full` is already what you'd expect from a TLS client to do during a handshake, i.e., verify the TLS server's host name.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /vendor
 /.vendor-new
 .vscode
+/.idea
```
Duologic, line 5: Do you mind adding this in your personal `$HOME/.gitignore`?

Makefile
```diff
@@ -100,4 +100,13 @@ crossplane.help:
 
 help-special: crossplane.help
 
+# NOTE(hasheddan): the build submodule currently overrides XDG_CACHE_HOME in
+# order to force the Helm 3 to use the .work/helm directory. This causes Go on
+# Linux machines to use that directory as the build cache as well. We should
+# adjust this behavior in the build submodule because it is also causing Linux
+# users to duplicate their build cache, but for now we just make it easier to
+# identify its location in CI so that we cache between builds.
+go.cachedir:
```
Duologic, line 109: I've incorporated this Make target in #88, do you mind rebasing?

Duologic:
Your arguments for changing the default seem valid to me.

apis/postgresql/v1alpha1/provider_types.go
Duologic: Should this be `*string`? (Kubernetes API guidelines require nil-able optional fields.)

.gitignore
ulucinar: Done.

Makefile
ulucinar: Done.

Duologic (approved):

ulucinar:
Hi @Duologic,
Thanks for the review! Looks like I do not have write access to the repo to merge this.

Duologic:
Sorry for the delay, I didn't want to merge it while on my phone in case I missed something.

zackb:
Hi, I have `sslMode: disable` in my definition and I am still encountering this bug. The ProviderConfig is deleted before the Database, Role, and Grant. Any help appreciated

PR #105, opened 2022-09-12
fix(integration): update mariaDB chart

Duologic:
### Description of your changes

The current chart does not seem to exist anymore (as reported in #97, this should fix
that.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

Run e2e test

[contribution process]: https://git.io/fj2m9

PR #118, opened 2022-11-10
Fix Grant revocation

### LINKED ISSUE #47: Deleting a MySQL User resource and then deleting a MySQL Grant resource linked to it causes the Grant deletion to hang
Grant deletion never occurs, and this warning repeats endlessly:
```
Warning  CannotObserveExternalResource  managed/grant.mysql.sql.crossplane.io  cannot show current grants: Error 1141: There is no such grant defined for user 'test-app' on host '%!'(MISSING)
```
This is on Crossplane 1.3.0 and MySQL 5.7. Can work on a repro case a bit later. The issue can be worked around by manually editing the Grant resource and removing the Crossplane finalizer.

jabbrwcky:
If the grant is already gone when the reconciler tries to revoke it, 
it should mark the resource as successfully removed, not error.

MySQL 8.x+ supports
`REVOKE %s IF EXISTS ON %s.%s FROM %s@%s IGNORE UNKNOWN USER` 
to avoid an error (which would be a cleaner solution IMO), 
but this will only be possible when dropping support for earlier versions.

Checks the actual mysql error code that is returned when the grant
revocation SQL errors; if the code states that the grant does not exist, 
it returns `nil` for successful removal of the grant.

Basically uses the same code as  https://github.com/crossplane-contrib/provider-sql/blob/ed6c55ce4798967e7652a4cce9f844b99ac1bbb7/pkg/controller/mysql/grant/reconciler.go#L227-L231

Fixes  #47

I have:

Added unit test for this case to `grants/reconciler_test.go` 

### DIFF
--- a/pkg/controller/mysql/grant/reconciler.go
+++ b/pkg/controller/mysql/grant/reconciler.go
@@ -328,6 +328,11 @@ func (c *external) Delete(ctx context.Context, mg resource.Managed) error {
 	)
 
 	if err := c.db.Exec(ctx, xsql.Query{String: query}); err != nil {
+		var myErr *mysqldriver.MySQLError
+		if errors.As(err, &myErr) && myErr.Number == errCodeNoSuchGrant {
+			//MySQL automatically deletes related grants if the user has been deleted
+			return nil
+		}
 		return errors.Wrap(err, errRevokeGrant)
 	}
 	err := c.db.Exec(ctx, xsql.Query{String: "FLUSH PRIVILEGES"})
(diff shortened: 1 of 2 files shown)


chlunde:
Nice, could you `git commit -s`? See https://github.com/crossplane-contrib/provider-sql/pull/118/checks?check_run_id=9415901160

 Just a heads up: I also think we will need a rebase when https://github.com/crossplane-contrib/provider-sql/pull/105 is merged

jabbrwcky:
Commit is signed

Duologic:
Can you fix the lint error?

jabbrwcky:
Fixed the reported linter issues.

You might consider upgrading golangci-lint (if possible for github-actions?), the output from v 1.50.1 is much more helpful:

```
pkg/controller/mysql/grant/reconciler.go:333:4: commentFormatting: put a space between `//` and comment text (gocritic)
                        //MySQL automatically deletes related grants if the user has been deleted
```

Duologic (approved):

PR #124, opened 2022-12-16
Issue #47 sceond attempt

### LINKED ISSUE #47: Deleting a MySQL User resource and then deleting a MySQL Grant resource linked to it causes the Grant deletion to hang

jabbrwcky:
Signed-off-by: Jens Hausherr <jens.hausherr@sapiens.com>

Caveat: At the moment this PR is primarily for discussing the approach. Test coverage et al. are not added yet.

After fixing failure to delete the next issue surfaced, thsi time the controller being unable to observe a deleted Grant.

IMO a better approach for the reconciler would be not to return an error if the Grant is missing, but to return a `managed.ExternalObservation{ResourceExists: false,ResourceUpToDate: false}`, which more accurately reflects the situation that the Grant does not exist and does not reflect the Grant resource.

Trying to update or to delete the resource can be left to the Create/Update/Delete methods of the controller.

Fixes #47

I have:

### DIFF
--- a/pkg/controller/mysql/grant/reconciler.go
+++ b/pkg/controller/mysql/grant/reconciler.go
@@ -145,6 +145,15 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 
 	privileges, result, err := c.getPrivileges(ctx, username, dbname, table)
 	if err != nil {
+		var myErr *mysqldriver.MySQLError
+		if errors.As(err, &myErr) && myErr.Number == errCodeNoSuchGrant {
+			// MySQL automatically deletes related grants if the user has been deleted
+			return managed.ExternalObservation{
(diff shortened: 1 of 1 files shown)


Duologic:
Did https://github.com/crossplane-contrib/provider-sql/pull/134 cover this?

Duologic:
Closing as stale.

Issue #130, opened 2023-02-14
SET session sql_log_bin = 0 when creating/updating/deleting user and grants in order not to break replication


alereca:
### What problem are you facing?

Creating different users for a master database and its replica would be great. Also, assigning different grants to users from a replica rather than the ones that are set in its master would be ideal for various use cases.

### How could Crossplane help solve your problem?

In https://github.com/crossplane-contrib/provider-sql/blob/master/pkg/controller/mysql/user/reconciler.go#L260 for example, before creating the user set the mysql session variable `sql_log_bin` to `0`:
```go
	if err := c.db.Exec(ctx, xsql.Query{
		String: "SET SESSION sql_log_bin = 0",
	}); err != nil {
		return managed.ExternalCreation{}, errors.Wrap(err, errFlushPriv)
	}
```
and then the other queries
```go
	query := fmt.Sprintf(
		"CREATE USER %s@%s IDENTIFIED BY %s%s",
		mysql.QuoteValue(username),
		mysql.QuoteValue(host),
		mysql.QuoteValue(pw),
		resourceOptions,
	)
	if err := c.db.Exec(ctx, xsql.Query{
		String: query,
	}); err != nil {
		return managed.ExternalCreation{}, errors.Wrap(err, errCreateUser)
	}
	if err := c.db.Exec(ctx, xsql.Query{
		String: "FLUSH PRIVILEGES",
	}); err != nil {
		return managed.ExternalCreation{}, errors.Wrap(err, errFlushPriv)
	}
```
I would like to work in a pr if this feature is considered as desirable

PR #135, opened 2023-03-07
Feat/queries with sql log bin to 0

alereca:

Fixes [#130](https://github.com/crossplane-contrib/provider-sql/issues/130)

I have:

Manually tested this, including an end-to-end test for this will involve deploying both a master and a replica database servers, I'm not sure whether that's desirable


apis/mysql/v1alpha1/user_types.go
```diff
@@ -45,6 +45,19 @@ type UserParameters struct {
 	// See https://dev.mysql.com/doc/refman/8.0/en/user-resources.html
 	// +optional
 	ResourceOptions *ResourceOptions `json:"resourceOptions,omitempty"`
+
+	// AuthPlugin sets the mysql authentication plugin, defaults to mysql_native_password
+	// +optional
+	// +kubebuilder:validation:Pattern:=^([a-z]+_)+[a-z]+$
+	AuthPlugin *string `json:"authPlugin,omitempty" default:"mysql_native_password"`
```
Duologic, line 52: I think AuthPlugin got in here accidentally?

pkg/controller/mysql/database/reconciler.go
```diff
@@ -149,8 +150,20 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
-	err := c.db.Exec(ctx, xsql.Query{String: "CREATE DATABASE " + mysql.QuoteIdentifier(meta.GetExternalName(cr))})
-	return managed.ExternalCreation{}, errors.Wrap(err, errCreateDB)
+	binlog := defaultBinlog(cr.Spec.ForProvider.BinLog)
+	if !binlog {
+		if err := c.db.Exec(ctx, xsql.Query{
+			String: "SET sql_log_bin = 0",
+		}); err != nil {
+			return managed.ExternalCreation{}, errors.Wrap(err, errSetSQLLogBin)
+		}
+	}
```
Duologic, line 160: It seems this code is repeated a few times, can it be abstracted to `pkg/clients/mysql` perhaps?

Duologic (commented):
Looks like an interesting feature.

apis/mysql/v1alpha1/user_types.go
alereca: Yes it did, thanks for pointing that out

alereca: Fixed by https://github.com/crossplane-contrib/provider-sql/pull/135/commits/d1320e0dd9b9e8d73973589e6ee21c5125735c95

pkg/controller/mysql/database/reconciler.go
alereca: Great idea! The only downside I see is that `func (c mySQLDB) Exec(ctx context.Context, q xsql.Query) error` would have to change is signature in order to access `cr.Spec.ForProvider.BinLog` boolean value. Is that refactor acceptable?

alereca: Since `func (c mySQLDB) Exec(ctx context.Context, q xsql.Query) error` is part of `xsql.DB` interface, which is common to mysql, mssql and postgres clients it would be very costly to refactor its signature. As every mock, call and implementation of that method should be updated. 

Consequently, I preferred another refactoring alternative to make it a bit less repetitive, creating a new function in `pkg/clients/mysql` called `ExecWithBinlogAndFlush` that is a wrapper function for `xsql.DB.Exec()` that allows disabling binlogs before the query and flushing privileges after the query (both optionally, as they default to true). 

What do you think? Would you consider this approach better than the previously mentioned refactor? I'm open to any further suggestions

Implemented in https://github.com/crossplane-contrib/provider-sql/pull/135/commits/0ba831f8ef0f2ec78b17f29b9957113bc11da51e

Duologic (approved):

alereca:
Resolved merge conflicts with master in https://github.com/crossplane-contrib/provider-sql/pull/135/commits/845f33f69680fc51276515d18161b35024fe6c96

Issue #92, opened 2022-05-12
postgresql - add support for "ALL" Privileges 


tilegg:
### What problem are you facing?
Grants for ALL privileges are not becoming ready

Example:
```
apiVersion: postgresql.sql.crossplane.io/v1alpha1
kind: Grant
metadata:
  name: example-role2-on-example-db
spec:
  forProvider:
    databaseRef:
      name: example-db
    roleRef:
      name: example-role
    privileges:
      - ALL
  providerConfigRef:
    name: postgres-example
```

### How could Crossplane help solve your problem?
The issue is due to postgresql basically interpolating ALL as CREATE, TEMPORARY, CONNECT (for databases). https://www.postgresql.org/docs/12/ddl-priv.html#PRIVILEGE-ABBREVS-TABLE

provider-sql expects the permissions to be ALL but the query returns false as they are "CREATE, TEMPORARY, CONNECT", therefore provider-sql tries to reconcile but never succeeds.

I've implemented a small PoC for databases (will link PR) that basically just mimics the interpolation postgres does for the observe function. If this approach makes sense to you it needs to be added to the other relevant resources. Also I was not able to figure out if and how to create a UnitTest for this.

tiegs:
Draft PR: https://github.com/crossplane-contrib/provider-sql/pull/93

PR #144, opened 2023-07-27
postgres: Expand grant privileges where required

iainlane:
Inside postgresql, some privileges are [altered][0] when issuing a `GRANT`, for example `ALL` is expanded to `CREATE, TEMPORARY, CONNECT`.

When we observe the grant, we query privileges on the database to see if the ones in the grant exist. Given that they are expanded, we can't simply query for what the user gave us. Expand the privileges which need expanding before we make the query.

Closes: #92

[0]: https://www.postgresql.org/docs/15/ddl-priv.html

  * There's a unit test to check the permissions are expanded
  * We extracted the queries the provider is using and ran them manually against a database

### DIFF
--- a/apis/postgresql/v1alpha1/grant_types.go
+++ b/apis/postgresql/v1alpha1/grant_types.go
@@ -43,6 +43,39 @@ type GrantPrivilege string
 // +kubebuilder:validation:MinItems:=1
 type GrantPrivileges []GrantPrivilege
 
+// Some privileges are shorthands for multiple privileges. These translations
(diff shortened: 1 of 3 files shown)


Duologic (approved):

iainlane:
Hmm, it works here 🤔 

image

I'll take a look later and see if CI does something different to me.

iainlane:
@Duologic I found the cause, nondeterministic map iteration order, should be fixed in e761eaa, could you take a look please?

PR #147, opened 2023-08-22
fix: remove unneeded flush privileges call for mysql #141 

### LINKED ISSUE #141: [mysql] FLUSH PRIVILEGES is only required after making manual edits to tables such as mysql.user, not after CREATE, ALTER or DROP USER

Why every time a `CREATE`, `ALTER` or `DROP USER` statement is executed, `FLUSH PRIVILEGES` is called after? It's also present after `GRANT` and `REVOKE` statements.

According to mysql documentation executing `FLUSH PRIVILEGES` is only required If you modify the grant tables directly using statements such as [INSERT](https://dev.mysql.com/doc/refman/8.0/en/insert.html), [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html), or [DELETE](https://dev.mysql.com/doc/refman/8.0/en/delete.html) (which is not recommended), the changes have no effect on privilege checking until you either tell the server to reload the tables or restart it. (https://dev.mysql.com/doc/refman/8.0/en/privilege-changes.html). Other references: 
- https://stackoverflow.com/questions/36463966/mysql-when-is-flush-privileges-in-mysql-really-needed
- https://docs.pingcap.com/tidb/dev/sql-statement-flush-privileges

Remove `FLUSH PRIVILEGES` statements from user and grant controllers, for example:
```go
		String: "FLUSH PRIVILEGES",
	}); err != nil {
		return managed.ExternalCreation{}, errors.Wrap(err, errFlushPriv)
	}
```
https://github.com/crossplane-contrib/provider-sql/blob/master/pkg/controller/mysql/user/reconciler.go#L271

I would like to work in a pr if this is considered as desirable

mariusziemke:

Fixes #141 

According to mysql documentation executing FLUSH PRIVILEGES is only required If you modify the grant tables directly using statements such as [INSERT](https://dev.mysql.com/doc/refman/8.0/en/insert.html), [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html), or [DELETE](https://dev.mysql.com/doc/refman/8.0/en/delete.html) (which is not recommended), the changes have no effect on privilege checking until you either tell the server to reload the tables or restart it. ([dev.mysql.com/doc/refman/8.0/en/privilege-changes.html](https://dev.mysql.com/doc/refman/8.0/en/privilege-changes.html)). Other references:

[stackoverflow.com/questions/36463966/mysql-when-is-flush-privileges-in-mysql-really-needed](https://stackoverflow.com/questions/36463966/mysql-when-is-flush-privileges-in-mysql-really-needed)
[docs.pingcap.com/tidb/dev/sql-statement-flush-privileges](https://docs.pingcap.com/tidb/dev/sql-statement-flush-privileges)

I have:

The change has been running in our various dev and production environments for a few weeks now. We tried to break our database clusters by adding and removing lots of users concurrently and everything works as expected. Before removing the Flush Privileges call we had a lot of trouble in our replicated db setup which tried to protect itself from inconsistency by just shutting down after adding more then one user at the same time.


mariusziemke:
Hi, can anyone look into this PR? Thanks!

Jibux:
We are interested on this fix because we have issue with our percona xtradb cluster witch can be stuck because of `FLUSH PRIVILEGES`. Thanks in advance for the reviewer(s).

fgaillot-qosenergy:
Also interested for having this merged for the next release, thanks in advance.

Duologic:
Would you mind rebasing?

mariusziemke:
@Duologic done.

Duologic:
Ah sorry, missed that this would conflict with #171, this might even simplify it more.

Duologic:
Merged in #202

PR #150, opened 2023-08-25
Support for SAP HANA database

adriantankodev:

My colleague @mirzakopic and I added support for the SAP HANA database. Following resources are available:

- Schemas
- Users
- Usergroups
- Roles

This of course allows other resources to be added later on.
We have also added only the basic resource parameters for now, but it is extendable.

I have:

In order to increase code isolation, we decided to separate controller and client logic. Therefore tests for both are also separated.


adriantankodev:
Unfortunately we messed up with merging/signing off. I opened a [new PR](https://github.com/crossplane-contrib/provider-sql/pull/151) from another branch.

PR #151, opened 2023-08-25
feat(hana): add support for hana

adriantankodev:

My colleague @mirzakopic and I added support for the SAP HANA database. Following resources are available:

- Schemas
- Users
- Usergroups
- Roles

This of course allows other resources to be added later on.
We have also added only the basic resource parameters for now, but it is extendable.

I have:

In order to increase code isolation, we decided to separate controller and client logic. Therefore tests for both are also separated.


Duologic:
My initial feeling is that this would fit better in separate provider (perhaps one dedicated and/or owned by SAP?), WDYT?

adriantankodev:
We have decided to create a separate provider as per your suggestion, so that it can be maintained directly by SAP. Thank you for the recommendation, I will go ahead and close the PR.

PR #171, opened 2024-02-08
Give MySQL sql_log_bin at DSN level to take parameter into account

### LINKED ISSUE #170: Disabling MySQL binlog writing does not work
### What happened?
Setting `binlog: false` to any of MySQL resources does not prevent MySQL server from writing to binlogs. Looks like this option is not applied at all.

### How can we reproduce it?

- Run a MySQL server with binlogs enabled
- Create a kube resource `User` for which `spec.forProvider.binlog = false`
- On MySQL server, run `mysqlbinlog -vv /var/lib/mysql/mysql-bin.XXXX`

→ You can see `CREATE USER` statement in binlogs

Crossplane version: v1.14.5
Provider-sql version: v0.9.0

### Analysis

By looking into the code, the `SET sql_log_bin = 0` statement is executed in a different MySQL connection from the one used for create/update/delete resource. In fact, as `mySQLDB.Exec()` creates a new `sql.Database` object, it also opens/closes a new every time it 's called, which explain why this option is not taken into account.

See [MySQL doc](https://dev.mysql.com/doc/refman/8.0/en/set-sql-log-bin.html) about this option.

I've been working on a patch to fix this, a PR will follow.

fgaillot-qosenergy:

Fixes #170

As actual MySQL code is not designed to execute several statements in the same connection, the `sql_log_bin` parameter value is now given at [DSN level](https://github.com/go-sql-driver/mysql/tree/v1.5.0?tab=readme-ov-file#system-variables). This ensures this parameter is always taken into account, no matter how MySQL connections are managed under the hood.

I guess this could be fixed differently, but it's the simpler way I found to manage this. Any feedbacks are welcome.

I have:

I created several `User` and `Grant` resources against a MySQL 8.0 server, with and without binlog option enabled, and then I check what was written in binlog files. It worked as expected.

### DIFF
--- a/pkg/clients/mysql/mysql.go
+++ b/pkg/clients/mysql/mysql.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/crossplane-contrib/provider-sql/pkg/clients/xsql"
@@ -16,7 +17,6 @@ import (
 
 const (
 	errNotSupported = "%s not supported by mysql client"
-	errSetSQLLogBin = "cannot set sql_log_bin = 0"
 	errFlushPriv    = "cannot flush privileges"
 )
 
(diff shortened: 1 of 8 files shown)


Duologic:
Would you mind rebasing?

fgaillot-qosenergy:
@Duologic Done

Duologic (approved):

PR #184, opened 2024-06-12
feat: add custom tls config support to mysql

silphid:

This PR adds support for custom TLS configuration to mysql implementation. In provider config file, if `tls` is set to `custom`, it reads custom TLS configuration from `tlsConfig` property, reading CA cert and client key/pair from K8s secret(s), and registering that config in mysql driver under the `custom` key.

Even though the mysql driver allows for multiple tls config key/value pairs, in the context of the provider it didn't appear to make sense to allow user to configure multiple TLS configurations and select only one of them, therefore the `tlsConfig` property is not a map, but rather a single config entry.

I have:

Because e2e tests require a totally different setup with a TLS-enabled mariaDB instance (but with same test cases), the current test script was duplicated and modified to add TLS, making sure that `make test-integration` runs both the no-tls and tls test scripts. It would be possible to refactor both scripts to combine them together and reduce duplication of setup and test code, however to the cost of readability. Let me know if that is a blocker and I will address it, I just didn't want to introduce more complexity in e2e test script until you confirm that's really what you prefer.


cluster/local/integration_tests_no_tls.sh
```diff
@@ -265,10 +266,10 @@ echo "${INSTALL_YAML}" | "${KUBECTL}" delete -f -
 timeout=60
 current=0
 step=3
-while [[ $(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l) != "0" ]]; do
+while [[ $(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l | tr -d '[:space:]') != "0" ]]; do
   echo "waiting for provider to be deleted for another $step seconds"
-  current=$current+$step
-  if ! [[ $timeout > $current ]]; then
+  current=$((current + step))
+  if [[ $current -ge $timeout ]]; then
```
silphid, line 272: This was also not ever timing out (only on macOS?), because doing integer operations/comparisons requires this special syntax.

cluster/local/integration_tests_tls.sh
```diff
@@ -0,0 +1,354 @@
...
+        secretName: mariadb-server-tls
+  extraVolumeMounts:
+    - name: tls-certificates
+      mountPath: /opt/bitnami/mariadb/certs
+      readOnly: true
+initdbScripts:
+  init.sql: |
+    CREATE USER 'test'@'%' IDENTIFIED BY '${mariadb_test_pw}' REQUIRE X509;
+    GRANT ALL PRIVILEGES ON *.* TO 'test'@'%' WITH GRANT OPTION;
+    FLUSH PRIVILEGES;
```
silphid, line 234: A dedicated `test` user is required in order to specifically require X509 on it, but not for `admin` user, which is also used for health probes without TLS.

cluster/local/integration_tests_tls.sh
```diff
@@ -0,0 +1,354 @@
...
+mariadb_root_pw=$(openssl rand -base64 32)
+mariadb_test_pw=$(openssl rand -base64 32)
+
+"${HELM3}" repo add bitnami https://charts.bitnami.com/bitnami
+
+MARIADB_VALUES=$(cat <<EOF
+auth:
+  rootPassword: ${mariadb_root_pw}
+primary:
+  extraFlags: "--ssl --require-secure-transport=ON --ssl-ca=/opt/bitnami/mariadb/certs/ca-cert.pem --ssl-cert=/opt/bitnami/mariadb/certs/server-cert.pem --ssl-key=/opt/bitnami/mariadb/certs/server-key.pem"
```
silphid, line 220: Note that `--require-secure-transport=ON` is not sufficient to require client to also provide its cert, we must also turn on `REQUIRE X509` on specific users (see below).

cluster/local/integration_tests_tls.sh
```diff
@@ -0,0 +1,354 @@
...
+      secretRef:
+        namespace: default
+        name: mariadb-creds
+        key: client-cert.pem
+    clientKey:
+      secretRef:
+        namespace: default
+        name: mariadb-creds
+        key: client-key.pem
+    insecureSkipVerify: true
```
silphid, line 282: `insecureSkipVerify` is to be used only here in e2e tests, because certs are self-signed, otherwise server would reject its own cert.

pkg/controller/mysql/database/reconciler.go
```diff
@@ -37,13 +37,15 @@ import (
...
 	"github.com/crossplane-contrib/provider-sql/pkg/clients/xsql"
+	"github.com/crossplane-contrib/provider-sql/pkg/controller/mysql/tls"
 )
 
 const (
 	errTrackPCUsage = "cannot track ProviderConfig usage"
 	errGetPC        = "cannot get ProviderConfig"
 	errNoSecretRef  = "ProviderConfig does not reference a credentials Secret"
 	errGetSecret    = "cannot get credentials Secret"
+	errTLSConfig    = "cannot load TLS config"
```
silphid, line 48: Used an error constant for consistency here, even though Crossplane contribution guidelines no longer recommend  using them.

cluster/local/integration_tests_tls.sh
```diff
@@ -0,0 +1,354 @@
...
+echo_info "check if connection secret has been updated"
+pw=$("${KUBECTL}" get secret example-connection-secret -ojsonpath='{.data.password}' | base64 --decode)
+[ "${pw}" == "${user_pw}" ]
+echo_step_completed
+
+echo_step "creating MySQL Grant resource"
+# create grant
+"${KUBECTL}" apply -f ${projectdir}/examples/mysql/grant_database.yaml
+
+echo_info "check if is ready"
```
ayahajar, line 329: check if what?

silphid:
@Duologic @chlunde @iainlane sorry for pinging you directly, I saw you seem to have been active here recently. Would just appreciate your cue on whether you think such a PR is likely to be reviewed on the short term, or if we should rather assume that we'll need to build and use our own fork for the next few months? Thanks! 🙏

cluster/local/integration_tests_no_tls.sh
```diff
@@ -265,10 +266,10 @@ echo "${INSTALL_YAML}" | "${KUBECTL}" delete -f -
 timeout=60
 current=0
 step=3
-while [[ $(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l) != "0" ]]; do
+while [[ $(kubectl get providerrevision.pkg.crossplane.io -o name | wc -l | tr -d '[:space:]') != "0" ]]; do
```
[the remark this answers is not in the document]
chlunde: probably also depending on PATH and brew packages, because I didn't notice anything on my mac :) But since this works on linux and in CI here I totally approve on the change :+1:

apis/mysql/v1alpha1/provider_types.go
```diff
@@ -31,9 +31,27 @@ type ProviderConfigSpec struct {
 	// or use preferred to use TLS only when advertised by the server. This is similar
 	// to skip-verify, but additionally allows a fallback to a connection which is
 	// not encrypted. Neither skip-verify nor preferred add any reliable security.
-	// +kubebuilder:validation:Enum="true";skip-verify;preferred
+	// Alternatively, set tls=custom and provide a custom TLS configuration via the tlsConfig field.
+	// +kubebuilder:validation:Enum="true";skip-verify;preferred;custom
 	// +optional
 	TLS *string `json:"tls"`
+
```
chlunde, line 38: you _may_ consider enforcing this
https://kubernetes.io/docs/reference/using-api/cel/

pkg/controller/mysql/tls/tls.go
```diff
@@ -0,0 +1,106 @@
...
+	if ok := pool.AppendCertsFromPEM(caCert); !ok {
+		return fmt.Errorf("cannot append CA certificate to pool")
+	}
+
+	keyPair, err := getClientKeyPair(ctx, kube, cfg)
+	if err != nil {
+		return err
+	}
+
+	return mysql.RegisterTLSConfig("custom", &tls.Config{
```
chlunde, line 67: Looks like "custom" is a global key here. I think this must have a unique name per DB if you connect to two databases?

pkg/controller/mysql/user/reconciler.go
```diff
@@ -115,6 +117,10 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 		return nil, errors.Wrap(err, errGetSecret)
 	}
 
+	if err := tls.LoadConfig(ctx, c.kube, pc.Spec.TLS, pc.Spec.TLSConfig); err != nil {
```
chlunde, line 120: Question: can and should this be called in newDB instead? It's something you always have to call before newDB? :thinking:

chlunde:
@silphid not sure if we should duplicate the full integration test, because there's also talk about adding PG integration tests.

Two ideas:
* Create a common library script with at least functions for  "setup cluster + install provider" and "teardown cluster", and reuse that code by importing it as a library using `source`
* Have a main driver script to run integration tests, calls smaller "integration_test_mysql_tls.sh", integration_test_mysql_no_tls.sh, integration_test_postgresql.sh, before cleanup. Each script can assume a working and "fairly clean" cluster. Each script must cleanup the cluster (but keep the provider installed).

@Duologic @Bastichou what do you think?

Duologic:
Thanks for this, I don't have any objections on either proposals, I gladly follow what ya'll think is a good path forward.

Thanks @chlunde for the review, I'm lacking bandwidth to review this but don't mind rubberstamping your approval and get this merged.

@silphid don't forget to sign the DCO ;)

apis/mysql/v1alpha1/provider_types.go
silphid: That would be an interesting exercise, but I have never used CEL, would need to read up on it and learn it, so I'd maybe hope to get away with it! 😅

pkg/controller/mysql/user/reconciler.go
silphid: We _could_ merge the two, but I see `newDB` as a non-failing function with external calls, no error result, etc, and I also perceive those two functions as conceptually different. We would need to refactor `newDB` considerably, along with corresponding unit tests, as it would slightly affect its scope/role. I tried to be as little intrusive as possible with my changes, but if we commonly decide that they should be merged together, I will tackle that refactoring.

silphid:
> @silphid not sure if we should duplicate the full integration test, because there's also talk about adding PG integration tests.

All right, I went ahead and refactored all integration tests into more modular shell functions. I kept them inline for now, as we only have mysql tests, but eventually if we add tests for PG, they would be easy to extract into other files can could be sourced. I'm pretty satisfied with the result, which is IMO much cleaner and readable. Let me know what you think! :)

silphid:
@chlunde @Duologic I just rebased on latest master, resolved conflicts and signed all commits (also see my above comments/improvements). Anything else I should do before you can run the workflow again?

silphid:
@chlunde I appreciate everyone is busy with their own projects, with little spare time left, but if it was possible to get a quick review on this one, I would be extremely grateful! 🙏☀️ We're hoping to start using those changes via the official release channel, but otherwise the plan is to setup the CI/CD for our fork internally, which I would sincerely prefer to avoid, if possible! 😅

silphid:
Bump @chlunde @Duologic.

Would so much appreciate your review and ideally unblocking the execution of the checks/workflows! 🙏

silphid:
Last bump @chlunde @Duologic 🤞 before we move on to setting up the build and publishing of our own fork, as a last resort.

Duologic (approved):

PR #186, opened 2024-07-01
Support RDS IAM authentication for MySQL

### LINKED ISSUE #106: Support RDS IAM authentication
Hi,

It seems that currently RDS IAM authentication for mysql/postgresql users is not possible with the provider.

Here are the official docs: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.DBAccounts.html

For mysql it is necessary to set `AWSAuthenticationPlugin` when creating the user.

For postgresql the user must be granted the `rds_iam` role.

This possible with the Terraform providers:
https://registry.terraform.io/providers/bangau1/mysql/latest/docs/resources/user
https://registry.terraform.io/providers/cyrilgdn/postgresql/latest/docs/resources/postgresql_grant_role

fortra-cloudops-platform:

Fixes #106, provides the ability to create MySQL users with the AWSAuthenticationPlugin

I have:

Honestly I have had issues testing as I don't know how to get the provider to run my image successfully, any help appreciated. 

### DIFF
--- a/apis/mysql/v1alpha1/user_types.go
+++ b/apis/mysql/v1alpha1/user_types.go
@@ -49,6 +49,11 @@ type UserParameters struct {
 	// BinLog defines whether the create, delete, update operations of this user are propagated to replicas. Defaults to true
 	// +optional
 	BinLog *bool `json:"binlog,omitempty" default:"true"`
+
+	// AuthPlugin defines the MySQL auth plugin (ie. AWSAuthenticationPlugin for AWS IAM authentication when using AWS RDS )
+	// See https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.DBAccounts.html
+	// +optional
+	AuthPlugin string `json:"authPlugin,omitempty" default:"mysql_native_password"`
 }
 
 // ResourceOptions define the account specific resource limits.
--- a/pkg/controller/mysql/user/reconciler.go
+++ b/pkg/controller/mysql/user/reconciler.go
@@ -54,6 +54,7 @@ const (
 	errUpdateUser              = "cannot update user"
 	errGetPasswordSecretFailed = "cannot get password secret"
 	errCompareResourceOptions  = "cannot compare desired and observed resource options"
+	errAuthPluginNotSupported  = "auth plugin not supported"
 
 	maxConcurrency = 5
 )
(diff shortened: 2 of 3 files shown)


mcanevet:
I was about to rebase #133, but it's a bit complicated because a lot of things have changed since that PR have been submitted.
So I tested that one and it works fine. Although, the default is `mysql_native_password` but this value throws `errAuthPluginNotSupported` error. Except form that it works fine.

cten:
I found that setting the default to mysql_native_password fails, documentation makes me believe it should work but failed on the MySQL version I tried. I can try to replicate the changes from #133 on current master. I think those might be cleaner code. Only difference would be not setting a default plugin name.

mcanevet:
At least I can confirm that this PR works fine. I'm using it right now.
Thanks a lot for this work!

mcanevet (approved):

pierluigilenoci:
@chlunde @Duologic, could someone from you review the PR?

pkg/controller/mysql/user/reconciler.go
```diff
@@ -238,20 +239,39 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+
+	if cr.Spec.ForProvider.AuthPlugin != "" {
+		authplugin = cr.Spec.ForProvider.AuthPlugin
 	}
+	var pw string
 
-	if pw == "" {
-		pw, err = password.Generate()
+	switch authplugin {
+	case "":
```
chlunde, line 253: default:"mysql_native_password"

is not handled here?

apis/mysql/v1alpha1/user_types.go
```diff
@@ -49,6 +49,11 @@ type UserParameters struct {
 	// BinLog defines whether the create, delete, update operations of this user are propagated to replicas. Defaults to true
 	// +optional
 	BinLog *bool `json:"binlog,omitempty" default:"true"`
+
+	// AuthPlugin defines the MySQL auth plugin (ie. AWSAuthenticationPlugin for AWS IAM authentication when using AWS RDS )
+	// See https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.DBAccounts.html
+	// +optional
+	AuthPlugin string `json:"authPlugin,omitempty" default:"mysql_native_password"`
```
chlunde, line 56: since the code only accepts specific values here, I think you should use an enum to limit to the valid types. Alternatively, let the code handle any value, if that possible

chlunde:
I'm not a maintainer of this project (and not a MySQL user), but as you pinged me I left some comments :)

A couple more:

1. this code needs unit tests so we cover the new branches

2. It _might_ be cleaner if you keep the query building part inside executeCreateUserQuery instead of partially building it outside, because now we get SQL quoting/escaping in different layers of the code

pierluigilenoci:
@fortra-cloudops-platform, can you check/react to PR comments?

Issue #110, opened 2022-10-27
Avoid reconciling many objects on the same database at the same time


chlunde:
### What problem are you facing?
For smaller RDS instances, we see CPU spikes when many provider-sql objects are reconciled at the same time. This can affect applications using the same database

### How could Crossplane help solve your problem?
Avoid reconciling all objects in a burst. For example:
* Connection reuse/pooling
* Max N objects per hostname reconciled at the same time
* Add jitter to spread reconciles

chkp-orhaz:
How do we avoid this?

chlunde:
@chkp-orhaz Looks like #197 will improve this!

PR #197, opened 2024-10-15
Provider-SQL reduce statement and connection overhead 

### LINKED ISSUE #194: Role provider should examine state of ConnectionLimit before applying

When a provider config contains a directive to set ConnectionLimit, it blindly applies the setting without checking the observed state as it does for Privileges and ConfigurationParameters. This is causing additional traffic to the catalog that is unhelpful in a distributed environment. 

In provider-sql/pkg/controller/postgresql/role/reconciler.go:
```
        cl := cr.Spec.ForProvider.ConnectionLimit
        if cl != nil && cr.Status.AtProvider.ConnectionLimit != cr.Spec.ForProvider.ConnectionLimit  { 
                        String: fmt.Sprintf("ALTER ROLE %s CONNECTION LIMIT %d", crn, int64(*cl)),
                }); err != nil {
                        return managed.ExternalUpdate{}, errors.Wrap(err, errUpdateRole)
                }
        }
```

### LINKED ISSUE #195: Provider should use connection pooling or single transaction to minimize load on database when reconciling

If I have hundreds to thousands of objects to be reconciled, opening a new connection for each aspect (Privs,Perms,Connections) creates unnecessary load to the database instance and can exhaust available connections. Let's say that I had dozens to 100's of roles to reconcile, I would end up creating for each user:

1 connection for CREATE ROLE xxxx PASSWORD yyyyyy ppppp
1 connection for ALTER ROLE xxxxx PASSWORD yyyyy
1 connection for ALTER ROLE xxxxxx pppppp
1 connection for ALTER  ROLE xxxxxx RESET ALL, and Configs in transaction block
1 connection for ALTER ROLE xxxxx CONNECTION LIMIT zzzzz

If the reconciler runs every 5 minutes, I would be churning thousands of connections per hour. 

Implement role provider with connection pooling (standard golang bindings are fine) and/or use ExecTx for all elements of each user.

ajcaldera1:

Fixes #194 
Fixes #195 

I have:

Local unit tests verified.
Scenario 1: Role created with no connection limit Result: Success, no further statements setting CONNECTION LIMIT were noted in the postgres.log 
Scenario 2: Role created with connection limit -1 Result: Success, no further statements setting CONNECTION LIMIT were noting in the postgres.log
Scenario 3: Role created with connection limit -1 and updated to connection limit 5 Result: Success, only 1 ALTER ROLE CONNECTION LIMIT statement was produced

Also noted reduced connection overhead with implementation of connection pooling.

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -24,13 +24,15 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	_ "github.com/lib/pq"
 
-	"gopkg.in/alecthomas/kingpin.v2"
+	"github.com/alecthomas/kingpin/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	xpcontroller "github.com/crossplane/crossplane-runtime/pkg/controller"
+	"github.com/crossplane/crossplane-runtime/pkg/feature"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
 
 	"github.com/crossplane-contrib/provider-sql/apis"
 	"github.com/crossplane-contrib/provider-sql/pkg/controller"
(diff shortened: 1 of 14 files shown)


chlunde (changes requested):
Very cool, I believe this also fixes #110

I think we should split out this into three PRs:
* ALTER ROLE CONNECTION LIMIT
* Connection pooling
* If required, any build system changes. There are some pending now so you might want to wait for them to be merged.

Could you
- fix DCO
- see if we could get test coverage for the changes?

PR #207, opened 2025-01-07
Feature/add revoke logic to grant pr1

olikyr:

This is a sub PR of 
https://github.com/crossplane-contrib/provider-sql/pull/200

1. I added a field called revokePublicOnDb in grant resource; if true, it will revoke from public access to DB; This usually a DBA requirement.
2. I added a field called ~~isPrivate~~ revokePublicOnSchema in schema resource; if true, it will revoke from public access to a schema; This usually a DBA requirement.

I have:

inside the project
```
make clean
make build
make reviewable
```


Duologic (commented):
Why are the parameters named differently?

olikyr:
I think you are right about this inconsistency, either they should be both isPrivate, or for schema adopt RevokePublicOnSchema.
I would go with RevokePublicOnSchema and RevokePublicOnDB as we are doing a specific command and I am afraid that private could infer a bigger context.
What do you think?

Duologic:
yeah, that's alright

Duologic (approved):

PR #210, opened 2025-01-15
chore: configure dependabot

Duologic:
This might be useful to keep dependencies up-to-date and prevent vulnerabilities to proliferate.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - directory: /
+    package-ecosystem: gomod
+    schedule:
+      interval: weekly
+
+  - directory: /
+    package-ecosystem: github-actions
+    schedule:
+      interval: weekly


benoittgt:
I think should be merged

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - directory: /
+    package-ecosystem: gomod
+    schedule:
+      interval: weekly
```
chlunde, line 6: monthly might be better for this project

And enable, if possible, "Dependabot security updates" too.

This will reduce the amount of dependabot churn but ensure timely security updates

chlunde: We might also want to do some grouping, for example Kubernetes dependencies.

chlunde: I've suggested an alternative here:

https://github.com/crossplane-contrib/provider-sql/pull/234

This will avoid creating PRs for bumping to a newer k8s api-version than crossplane-runtime has tested, and also avoid having the history of the project just be lots of dependabot updates without releases.

PR #216, opened 2025-02-18
chore: bump golangci version

fernandezcuesta:
`make lint` command prompts deprecation warnings and fails to succeed with the current settings, mainly due to deprecated linters. This PR bumps golangci version and replaces some linters with the suggested alternatives, while sticking as much as possible to the original settings.

Fixes #

I have:

`make lint`

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -18,11 +18,11 @@ jobs:
   # The main gotcha with this action is that PRs _must_ be labelled before they're
   # merged to trigger a backport.
   open-pr:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     if: github.event.pull_request.merged
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
       - name: Open Backport PR
         uses: korthout/backport-action@v1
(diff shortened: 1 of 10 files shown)


Duologic (changes requested):
Can you also update it here: https://github.com/crossplane-contrib/provider-sql/blob/master/.github/workflows/ci.yml#L14

fernandezcuesta:
> Can you also update it here: https://github.com/crossplane-contrib/provider-sql/blob/master/.github/workflows/ci.yml#L14

Done, thanks!

Duologic (approved):

fernandezcuesta:
Hi @Duologic I see this is failing due to some deprecated actions (https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). WDYT about using this very same PR to bump the actions?

BTW just saw [the dependabot PR](https://github.com/crossplane-contrib/provider-sql/pull/210) which would have helped raising the flag here.

chlunde:
@fernandezcuesta If you can add the publish-artifact bump to this PR, that would be great!

chlunde (approved):

chlunde:
@fernandezcuesta for some reason, we don't get any button here to approve workflows from first time contributor in this PR. Could you try and create a new PR, as it seems you already tried rebase and that didn't help.

PR #237, opened 2025-05-23
Suppress unnecessary changes to connection limits by checking existing value for PostgreSQL

### LINKED ISSUE #194: Role provider should examine state of ConnectionLimit before applying

ajcaldera1:
This change checks the state of the connection limit for a Postgres role prior to mutating it. The previous behavior would blindly fire the ALTER ROLE <role> CONNECTION LIMIT -1 even if there was no change. Added test to verify connection limit settings were being properly mutated. Corrected a couple of instances of `kubectl` being referenced in the tests that caused them to fail, replaced with shell variable `${KUBECTL}`. 

Fixes #194

I have:

Added test to e2e suite to verify the connection limit and review Postgres logs to ensure that no statement had been fired when it was unnecessary. 


pkg/controller/postgresql/role/reconciler.go
```diff
@@ -405,10 +408,11 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 		// Update state to reflect the current configuration parameters
 		cr.Status.AtProvider.ConfigurationParameters = cr.Spec.ForProvider.ConfigurationParameters
 	}
-	cl := cr.Spec.ForProvider.ConnectionLimit
-	if cl != nil {
+	newCl := cr.Spec.ForProvider.ConnectionLimit
+	currCl := cr.Status.AtProvider.ConnectionLimit
+	if (newCl != nil && currCl != nil) && (int64(*currCl) != int64(*newCl)) {
```
chlunde, line 413: Why the three casts to int64? Could we simply drop then?

chlunde:
I'm closing this too as there has been no follow up with this and other PRs like #434 and #453 touches on the same issues. We'll attribute you in the release notes, thanks again for working on this!

PR #238, opened 2025-05-27
feat(postgresql): add TLS related options

zregvart:

This adds `sslCert`, `sslKey` and `sslRootCert` options to the PostgreSQL `ProviderConfig` needed to establish a connection to a database that either uses non-public certificates or requires client certificate authentication.

I've opted to create the `Options` struct to hold these and the existing `sslMode` parameter, this will allow adding other options[1] easier in the future.

[1] https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

I have:

Added a simple unit test

### DIFF
--- a/cluster/local/postgresdb_functions.sh
+++ b/cluster/local/postgresdb_functions.sh
@@ -137,6 +137,33 @@ EOF
   "${KUBECTL}" patch providers.pkg.crossplane.io/provider-sql --type=json --patch='[{"op":"add","path":"/spec/runtimeConfigRef","value":{"name":"postgres-tls"}}]'
 }
 
+setup_provider_config_postgres_inline_tls() {
+  echo_step "creating ProviderConfig for PostgresDb with inline TLS"
+  local yaml="$(cat <<EOF
+---
+apiVersion: postgresql.sql.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: default
+spec:
+  credentials:
+    source: PostgreSQLConnectionSecret
+    connectionSecretRef:
+      namespace: default
+      name: postgresdb-creds
+EOF
+  )"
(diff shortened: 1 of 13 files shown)


zregvart:
Thanks for having a look. Yes, this would require mounting files to the provider container, this was simplest for me to do. I can also, or solely, use a Secret, the secret would need to be fetched and placed in the $TMPDIR as well, or, I think, the `sslinline=true` option could be used to provide the certificate and keys inline.

I had a stab at the e2e tests, and I encountered a number of issues with port forwarding, not starting from the clean slate, etc. I have a work in progress commit [here](https://github.com/zregvart/provider-sql/commit/9c60a76532639d211b0a17d81f0ad6fbd9eb1153). Have a look at how the CA certificate from secret auto-generated by the Helm chart for PostgreSQL is mounted.

zregvart:
@chlunde I've added the e2e tests in 3ff79528f5e539e3b0b2b86b95842a6e8fa73977 and an example in 2e971b53c1df2d42649cbe21d303ad5b55dd5ebc, when you get a chance I would appreciate if you could take another look.

The GitHub Discussions don't appear to be enabled, or I can't access them.

chlunde:
@zregvart cool, I do think we should look at secrets. For the public key it might make sense with a file but for client certs that might be per server or even role, right? I'll have to research a bit here and come back to you.

zregvart:
0f2afad62eb292dcbf1332b0b17a72bfdfeb8a17 adds reading the Secret in addition to the changes made to the ProviderConfig previously. I needed to place the data from the Secret in a temporary file, the `sslinline=true` was not helpful the values were broken up on the newlines of the PEM encoded certificates/keys and the single quoting the value did not help as the value was escaped and had no impact.

@chlunde when you get a chance I'd be happy to receive any feedback on this

PR #301, opened 2025-12-29
feat: add cache sync timeout configuration for controller manager

fahedouch:

Added a configurable cache synchronization timeout to avoid error : 
```
Informer fails to sync, timeout waiting for API
```

I have:


chlunde:
I can't see any other controllers with this exact string:

https://github.com/search?q=org%3Acrossplane-contrib+%22cache-sync-timeout%22&type=code

or CacheSyncTimeout

Did you have to do this in crossplane and providers too?

I think we should have the same option for all providers if we need this, but I'm curious why you only hit it on provider-sql.

chlunde:
@fahedouch if you get it into provider-template then we'll get it in here! Thanks

PR #343, opened 2026-03-12
feat(postgres): Add support for setting STRATEGY during db creation

alexdor:

Add PostgreSQL Database strategy support for template-based database creation and make its lifecycle explicit in the API. The new strategy field is treated as create-time-only, and validated so updates do not silently drift from the real database state.

Fail fast when strategy is requested against PostgreSQL versions older than 15 instead of letting CREATE DATABASE fail with a less clear SQL error.

Fixes #341

I have:

- run test-integration
- local testing


alexdor:
@fernandezcuesta sorry I thought that it was a flaky test (because it would pass locally sometimes, and I didn't touch any of the mysql parts). Should be fixed now

chlunde (approved):
Thanks!

Issue #297, opened 2025-12-04
Question: how to rotate MySQL user passwords


lah-wag:
Hi,

I have some databases, users and grant created with the provider. The passwordRef field is empty, so passwords are auto-generated.
How can I trigger a rotation of the passwords?

I'm running version 0.9

/Lasse

PR #348, opened 2026-03-23
feat(postgresql): auto-regenerate role password on restore and support rotation trigger

### LINKED ISSUE #227: Generate a new PostgreSQL password for roles in case of restoration
Hi provider-sql maintainers,

When I recreate an **orphaned role** or in case of cluster restoration, if I haven't backed up the secrets containing the PostgreSQL passwords of the roles, I am unable to reuse these roles for my deployments.

I saw that the problem is due to the encryption of the passwords into the database so we can't retrieve them after creation.

It could be great to implement a behaviour like in the Postgres-operator (if not already done) that generate a new password and replaces the previous one into the database so there is no need to keep the old secrets in a safe location to be able to restore the user later.

Thank you for your help

JevgeniF:

Fixes #227: The change adds an optional boolean field, `passwordReset`, to the Role spec. This field changes how the provider behaves when it encounters an empty Secret resource after a restoration:

If `passwordReset` is true and the Secret is empty, the provider generates a new password for the role and populates the Secret.

If `passwordReset` is false or omitted, the provider acts as usual and leaves the role's password unchanged (meaning the Secret may simply stay empty).

I have:

Unit tests written to test passwordResetToken work.
Manual test in AWS cluster.

A test linux_amd64 package for you reference can be found in Docker Hub:
`docker pull entigolabs/provider-sql:test`
Package digest: `sha256:6b41c2bc13fc23ecd13f9461d3beeb3d38af3990c6d3c1b75167ad0260382d05`


fernandezcuesta:
@JevgeniF thanks for your contribution!

I like the idea, wondering if it can be extended to become a sort of either:

- autogenerate password
- or bring your own password, in which case the connection secret is redundant

I did some related work in another provider for a User resource in [provider-clickhousedbops](https://github.com/lansweeper-oss/provider-clickhousedbops/releases/tag/v0.2.0)

JevgeniF:
Dear @fernandezcuesta,

Your idea to bring BYOP to this project is great, but I’m not sure it should be solved within the context of this specific fix.
The goal of this fix is to allow users to use an already existing PostgreSQL Role (one restored from a snapshot). Currently, we have to delete the old user and create a new one just to get a properly populated Secret, which is why we need this fix so urgently.
Maybe the BYOP feature can be moved into a separate ticket? I’d be happy to see what I can do there.

apis/cluster/postgresql/v1alpha1/role_types.go
```diff
@@ -90,6 +90,12 @@ type RoleParameters struct {
 	// See https://www.postgresql.org/docs/current/runtime-config-client.html for some available configuration parameters.
 	// +optional
 	ConfigurationParameters *[]RoleConfigurationParameter `json:"configurationParameters,omitempty"`
+
+	// PasswordReset controls behaviour when the role exists in the database but the connection
+	// secret has no password.
```
fernandezcuesta, line 95:
```suggestion
	// secret has no password. Has no effect when passwordSecretRef is set.
```

apis/namespaced/postgresql/v1alpha1/role_types.go
```diff
@@ -91,6 +91,12 @@ type RoleParameters struct {
 	// See https://www.postgresql.org/docs/current/runtime-config-client.html for some available configuration parameters.
 	// +optional
 	ConfigurationParameters *[]RoleConfigurationParameter `json:"configurationParameters,omitempty"`
+
+	// PasswordReset controls behaviour when the role exists in the database but the connection
+	// secret has no password.
```
fernandezcuesta, line 96:
```suggestion
	// secret has no password. Has no effect when passwordSecretRef is set.
```

pkg/controller/cluster/postgresql/role/reconciler.go
fernandezcuesta, line 275:
```suggestion
passwordResetNeeded := c.shouldResetPassword(mg)

// PrivilegesAsClauses is used as role status output
mg.Status.AtProvider.PrivilegesAsClauses = privilegesToClauses(observed.Privileges)

return managed.ExternalObservation{
ResourceExists:          true,
ResourceLateInitialized: lateInit(observed, &mg.Spec.ForProvider),
ResourceUpToDate:        !pwdChanged && !passwordResetNeeded && upToDate(observed, &mg.Spec.ForProvider),
```

fernandezcuesta, line 340:
```suggestion

if !pwchanged && c.shouldResetPassword(mg) {
	pw, err = password.Generate()
	if err != nil {
		return managed.ExternalUpdate{}, err
	}
	pwdchanged = true
}

```

pkg/controller/cluster/postgresql/role/reconciler.go
```diff
@@ -341,6 +341,12 @@ func (c *external) Update(ctx context.Context, mg *v1alpha1.Role) (managed.Exter
 	crn := pq.QuoteIdentifier(meta.GetExternalName(mg))
 
 	if pwchanged {
+		if pw == "" {
+			pw, err = password.Generate()
+			if err != nil {
+				return managed.ExternalUpdate{}, err
+			}
+		}
```
fernandezcuesta, line 349:
```suggestion
```

pkg/controller/cluster/postgresql/role/reconciler.go
fernandezcuesta, line 354:
```suggestion
		}
		now := metav1.Now()
		mg.Status.AtProvider.LastPasswordChange = &now
```

pkg/controller/cluster/postgresql/role/utils.go
```diff
@@ -31,36 +31,59 @@ import (
...
-		return "", false, nil
-	}
-	nn := types.NamespacedName{
-		Name:      role.Spec.ForProvider.PasswordSecretRef.Name,
-		Namespace: role.Spec.ForProvider.PasswordSecretRef.Namespace,
-	}
-	s := &corev1.Secret{}
-	if err := c.kube.Get(ctx, nn, s); err != nil {
-		return "", false, errors.Wrap(err, errGetPasswordSecretFailed)
+	if role.Spec.ForProvider.PasswordSecretRef != nil {
```
fernandezcuesta, line 34: I think you don't need to change any of this

pkg/controller/cluster/postgresql/role/utils.go
```diff
@@ -31,36 +31,59 @@ import (
...
-		return "", false, err
+		return false, err
 	}
-	// if newPwd was set to some value, compare value in output secret with
-	// newPwd
-	changed = newPwd != "" && newPwd != string(s.Data[xpv1.ResourceCredentialsSecretPasswordKey])
-
-	return newPwd, changed, nil
+	return len(s.Data[xpv1.ResourceCredentialsSecretPasswordKey]) == 0, nil
 }
```
fernandezcuesta, line 89:
```suggestion
func (c *external) shouldResetPassword(ctx context.Context, role *v1alpha1.Role) bool {
	if role.Spec.ForProvider.PasswordSecretRef != nil {
		return false
	}
	trigger := role.Spec.ForProvider.PasswordReset
	if trigger == nil {
		return false
	}
	last := role.Status.AtProvider.LastPasswordChange
	if last == nil {
		return true
	}
	return trigger.After(last.Time)
}
```

fernandezcuesta (commented):
Alternative implementation, considering that:

- passwordSecretRef implements BYOP (and this trigger should have no effect)
- Set a timestamp indicating when you want the password to be reset, so it somehow changes the logic of "reset the password"


(missing: update in role_types to add PasswordRotationTrigger in roleparameters and LastPasswordChange in RoleObservation + tests)

:warning: take that just as an idea, I'm not pretending you to change that, but just bring more ideas.

pkg/controller/cluster/postgresql/role/utils.go
JevgeniF: Kept `PasswordReset` as `*bool`, because it should fire once when `LastPasswordChange `is nil (meaning the provider has never set a password for this role), not on a schedule (as in our issue, when we restore database with roles, but provider doesn't generate password).

For time-based rotation I added a separate `PasswordRotationTrigger *metav1.Time` field, which does use `trigger.After(last.Time)` in `shouldResetPassword`. So both use cases are covered:          
  -- `PasswordReset: true` - fires once when `LastPasswordChange == nil`
  -- `PasswordRotationTrigger` - fires whenever the trigger time is after `LastPasswordChange`

pkg/controller/cluster/postgresql/role/reconciler.go
JevgeniF: I didn't add this block separately for the same reason as before: `getPassword` already calls `shouldResetPassword` for the non-BYOP case and returns its result as changed.

JevgeniF: I didn't add an `passwordResetNeeded` variable in `ExternalObservation`, because `getPassword` already calls `shouldResetPassword` internally for the non-BYOP cases and returns its result as changed.
So  `!pwdChanged` in `ResourceUpToDate` already covers both cases:                                                                                                                                      
-- BYOP case: `pwdChanged` = password diff between user secret and connection secret
-- Non-BYOP path: `pwdChanged` = shouldResetPassword(role)                                                                                                                                         

For BYOP, `shouldResetPassword` always returns false (guarded by `passwordSecretRef != nil`), so a separate variable would add no information.

JevgeniF:
@fernandezcuesta maybe I got your idea wrong and made wrong implementation. Please look again and build and test if BYOP and password rotation works as you wanted.

New field added to `RoleParameters`:                                                                                                                                                     
 -- `PasswordReset *bool`                                                                                                                                                    
-- `PasswordRotationTrigger *metav1.Time` - rotation trigger for auto-generated passwords. Fires whenever the trigger time is after LastPasswordChange.                                   

Both `PasswordReset` and `PasswordRotationTrigger` have no effect when `passwordSecretRef `is set - BYOP rotation behavior unchanged.                                                                                                                                                                                         

A new `LastPasswordChange *metav1.Time` status field is added to `RoleObservation`. It is set by the provider after every successful password change (both on create and update) and is used by `shouldResetPassword` to decide whether action is needed.

`shouldResetPassword` is refactored from `(ctx, role) (bool, error)` to `(role) bool` - no kube call required since the decision is based on fields already present in the role object.

fernandezcuesta:
Hey, thanks for considering it and your explanation. I'm wondering if both fields aren't a bit redundant.  I mean, with a timestamp-alike field, can't you satisfy the scenario in which you lost the password and want to have it immediately recreated? In the case of a "past" time, wouldn't it act similar to PasswordReset?

JevgeniF:
@fernandezcuesta you are totally right. I was overthinking and left our simple (legacy) logic in parallel with your solution, but `PasswordRotationTrigger` indeed covers both cases. Furthermore, If we set `LastPasswordChange` in Create() method, we can trigger password change after database restoration by default, without `PasswordRotationTrigger` use. 

To sum up...
**Current logic for password rotation (BYOP - `PasswordSecretRef` exists):**

1.  The provider generates its own connection secret based on the referenced user secret (note: `writeConnectionSecretToRef` must be configured).
2. When a user rotates the password in the source secret, the provider detects the difference between the source and the managed connection secret and triggers an update.

Am I right?

**Updated logic for non-BYOP (`PasswordSecretRef `not set), considering my proposition to keep it simple on `LastPasswordChange` basis.**

**For new role creation:**

1. The provider generates a secret using `Create()` function and sets the `LastPasswordChange` timestamp in the resource `status`.
2. During subsequent observations, `shouldResetPassword` returns false if `PasswordRotationTrigger` is either nil or set to a timestamp in future in comparison with `LastPasswordChange`. It returns true only if `PasswordRotationTrigger `is older than `LastPasswordChange`.
3. if `PasswordRotationTrigger` is older than `LastPasswordChange`, the provider updates password and sets new `LastPasswordChange` value.

**For restoration scenarios:**

1. The provider observes that the role already exists in the database and bypasses the `Create() `function. Consequently, `LastPasswordChange` remains nil in the new resource status.
2. A nil value for `LastPasswordChange` forces `shouldResetPassword` to true, prompting the provider to call the `Update()` function.
3. The provider updates the database password, populates the secret and sets `LastPasswordChange` .
4. From this point, `PasswordRotationTrigger` functions normally, triggering a change only when its timestamp is more older than `LastPasswordChange`.

Looks simplier and works in any scenario:
provider will automatically change password for restored role.
user can reset password manually by changing `PasswordRotationTrigger`

fernandezcuesta:
Yes I guess that's correct

JevgeniF:
@fernandezcuesta OK. Then I push updated version tomorrow. Thank you!

pkg/controller/cluster/postgresql/role/reconciler.go
JevgeniF: Returned it back, because now - i.e., when we need to reset a password in the non-BYOP case - `c.getPassword` returns `"", true, nil`, and we need to generate the password before the `ALTER ROLE ... PASSWORD ...` query in the database."

JevgeniF:
Some comments.
I made password reset by default in case if role created and password in connection secret remains empty.
I also removed `lastPasswordChange` update from`Create() `function (which I added in previous version). Now the field populated only on password change (on `Update()`)

**Current flows:**
**Normal role creation (non-BYOP):** 
Role doesn't exist in DB yet. Provider creates it with a generated password and publishes the connection secret. On the next reconcile, `lastPasswordChange` is nil but the connection secret already has a password - provider knows `Create()` already ran and leaves it alone. 

**Role restoration (non-BYOP):**
Role already exists in DB (restored from snapshot), but the `Role` resource is new so there's no connection secret or it is empty. Provider sees `lastPasswordChange` nil and empty secret - provider gets this as trigger for password generation - generates a new one, sets it on the role in DB via `ALTER ROLE`, writes the connection secret, stamps `lastPasswordChange`.   

**Password rotation (non-BYOP):** 
`passwordRotationTrigger`: triggers new password generation on comparison with` lastPasswordChange` basis.

New inux_amd64 test package:
`docker pull entigolabs/platform-apis-test:provider-sql-test-v4`
Package digest: `sha256:8cc37ccf50c6071587dff2b01da0dcd003637982f63c4a0ea1bc0e8f95ddb983`

JevgeniF:
Dear @fernandezcuesta,

Hope you are doing well. What our next steps with this PR?

fernandezcuesta:
Hi @JevgeniF, I'm thinking that probably other resources which may benefit from the same improvement are MySQL and MSSQL User. WDYT to extend/standardize the logic to also cover those? Would also close #297

JevgeniF:
Hi @fernandezcuesta,

Thanks for the suggestion! Extending this pattern to MySQL and MSSQL makes perfect sense from a standardization perspective.

However, since our current cloud infrastructure only uses PostgreSQL, implementing and testing this across other databases would require time and resources we unfortunately can't spare right now. Getting this feature live for PostgreSQL is also a high priority for our team.

I’d love to see this standardized eventually, but for now, I can only commit to the PostgreSQL fix for ticket 227. I might look into the other databases as a separate project down the road, but for this specific PR, I need to keep the scope limited to PostgreSQL so we can get it merged as soon as possible.

Thanks for understanding!

fernandezcuesta:
Sorry for the delay, let's merge this PR as is, and open a new issue to track the generalization to other DBs.

fernandezcuesta (approved):

fernandezcuesta:
Thanks @JevgeniF for your contribution :tada: :rocket:

fernandezcuesta:
... and a very last minute comment:

`utils.go:86`: `kube.Get` errors (including transient network failures) are silently ignored:

```go
  if err := c.kube.Get(ctx, nn, s); err == nil {
```

:arrow_forward: Any non-nil error = treated as _"secret has no password"_ = triggers password reset. A transient API error could cause an unnecessary password rotation. 

Consider at minimum logging the error, or distinguishing `NotFound` from other errors.

JevgeniF:
@fernandezcuesta Sorry for the delay from my side. Silent error of errors in shouldResetPassword fixed.

pkg/controller/cluster/postgresql/role/utils.go
```diff
@@ -31,36 +31,73 @@ import (
...
-		Namespace: role.Spec.WriteConnectionSecretToReference.Namespace,
+	shouldReset, err := c.shouldResetPassword(ctx, role)
+	return "", shouldReset, err
+}
+
+// shouldResetPassword returns true when a password change is needed for the non-BYOP path.
+// When LastPasswordChange is nil: checks the connection secret — if it already has a
+// password, no reset is needed; if the secret is
+// absent or empty, a role restoration is assumed and a reset is triggered.
+// Also returns true when PasswordRotationTrigger is set to a time after LastPasswordChange.
```
fernandezcuesta:
```suggestion
// shouldResetPassword returns true when a password change is needed for the non-BYOP path.
// When LastPasswordChange is set, only a PasswordRotationTrigger newer than it forces a
// reset. When LastPasswordChange is nil the connection secret decides: a populated secret
// means Create already ran, while an absent or empty secret means the role was restored
// out-of-band and needs a fresh password. Without a connection secret reference there is
// nowhere to publish a regenerated password, so no reset is attempted.
```

pkg/controller/cluster/postgresql/role/utils.go
```diff
@@ -31,36 +31,73 @@ import (
...
+				if resource.IgnoreNotFound(err) != nil {
+					return false, err
+				}
+				return true, nil
+			}
+			if len(s.Data[xpv1.ResourceCredentialsSecretPasswordKey]) > 0 {
+				return false, nil
+			}
+		}
+		return true, nil
```
fernandezcuesta:
```suggestion
     if last != nil {
             if role.Spec.ForProvider.PasswordRotationTrigger != nil {
                     return role.Spec.ForProvider.PasswordRotationTrigger.After(last.Time), nil
             }
             return false, nil
     }
```

pkg/controller/cluster/postgresql/role/utils.go
```diff
@@ -31,36 +31,73 @@ import (
...
-		return "", false, err
+	if role.Spec.ForProvider.PasswordRotationTrigger != nil {
+		return role.Spec.ForProvider.PasswordRotationTrigger.After(last.Time), nil
 	}
-	// if newPwd was set to some value, compare value in output secret with
-	// newPwd
-	changed = newPwd != "" && newPwd != string(s.Data[xpv1.ResourceCredentialsSecretPasswordKey])
-
-	return newPwd, changed, nil
+	return false, nil
```
fernandezcuesta:
```suggestion
     if role.Spec.WriteConnectionSecretToReference == nil {
             return false, nil
     }
     nn := types.NamespacedName{
             Name:      role.Spec.WriteConnectionSecretToReference.Name,
             Namespace: role.Spec.WriteConnectionSecretToReference.Namespace,
     }
     s := &corev1.Secret{}
     err := c.kube.Get(ctx, nn, s)
     if err != nil {
             if resource.IgnoreNotFound(err) != nil {
                     return false, errors.Wrap(err, errGetConnectionSecretFailed)
             }
             return true, nil
     }
     return len(s.Data[xpv1.ResourceCredentialsSecretPasswordKey]) == 0, nil
```

fernandezcuesta: Plus new const in `reconciler.go`:

```go
     errGetConnectionSecretFailed = "cannot get connection secret"
```

fernandezcuesta: lastPasswordChange |     trigger     | secret ref | secret state  |                                  reset                                  |
-------------------|-----------------|------------|---------------|-------------------------------------------------------------------------|
set                | > last          | any        | any           | yes (rotation)                                                          |
set                | <= last / nil   | any        | any           | no                                                                      |
nil                | set (any value) | any        | any           | yes - bug: skips secret check, spurious rotation post-Create            |
nil                | nil             | yes        | missing/empty | yes (restore, intended fix)                                             |
nil                | nil             | yes        | has password  | no (post-Create guard)                                                  |
nil                | nil             | no         |               | yes - bug: upgrade rotates secretless roles, new password unpublishable |

Row 3: trigger presence alone forces reset whenever status nil, bypassing populated-secret guard
Row 6: no-secret-ref fallthrough hits final return true, nil.

fernandezcuesta: Also, take into account expectation for test `NilLastPasswordChange` (no secret ref, no trigger): `changed: true -> false` (flip expectation).

fernandezcuesta: And optionally add a couple of cases for `TestGetPassword` (vibe coded, didn't check):

```go
"RotationTriggerNilLastChangeSecretHasPassword": {
      reason: "PasswordRotationTrigger with nil LastPasswordChange must not reset when the connection secret already has a password (Create just ran)",
      args: args{
              role: &v1alpha1.Role{
                      Spec: v1alpha1.RoleSpec{
                              ForProvider: v1alpha1.RoleParameters{
                                      PasswordRotationTrigger: &v1.Time{Time: time.Now()},
                              },
                              ResourceSpec: xpv1.ResourceSpec{
                                      WriteConnectionSecretToReference: &xpv1.SecretReference{
                                              Name:      "test-secret",
                                              Namespace: "test-ns",
                                      },
                              },
                      },
              },
              kube: &test.MockClient{
                      MockGet: func(_ context.Context, _ client.ObjectKey, obj client.Object) error {
                              secret := corev1.Secret{
                                      Data: map[string][]byte{
                                              xpv1.ResourceCredentialsSecretPasswordKey: []byte("existing-password"),
                                      },
                              }
                              secret.DeepCopyInto(obj.(*corev1.Secret))
                              return nil
                      },
              },
      },
      want: want{pwd: "", changed: false},
},
"NilLastPasswordChangeSecretNotFound": {
      reason: "nil LastPasswordChange with a missing connection secret should trigger a reset (restoration scenario)",
      args: args{
              role: &v1alpha1.Role{
                      Spec: v1alpha1.RoleSpec{
                              ResourceSpec: xpv1.ResourceSpec{
                                      WriteConnectionSecretToReference: &xpv1.SecretReference{
                                              Name:      "test-secret",
                                              Namespace: "test-ns",
                                      },
                              },
                      },
              },
              kube: &test.MockClient{
                      MockGet: test.NewMockGetFn(kerrors.NewNotFound(schema.GroupResource{Resource: "secrets"}, "test-secret")),
              },
      },
      want: want{pwd: "", changed: true},
},
```

fernandezcuesta (changes requested):

fernandezcuesta:
Sorry @JevgeniF last minute bug found

pkg/controller/cluster/postgresql/role/utils.go
JevgeniF: added

fernandezcuesta (approved):

fernandezcuesta:
Thanks again @JevgeniF, LGTM :rocket:

PR #357, opened 2026-04-09
feat: tunnel raw TCP database connections through HTTP CONNECT proxy

jaymiracola:
## Summary

- Adds HTTP CONNECT proxy support to PostgreSQL, MySQL, and MSSQL database drivers
- When `HTTP_PROXY` or `HTTPS_PROXY` is set, connections are tunnelled through the proxy via HTTP CONNECT rather than dialling the database endpoint directly
- Mirrors the behaviour that `net/http`-based providers get automatically, extending it to raw TCP database drivers that bypass `net/http` entirely

## Background

Database drivers (`lib/pq`, `go-sql-driver/mysql`, `go-mssqldb`) use `net.Dial` directly and do not respect `HTTP_PROXY`/`HTTPS_PROXY` environment variables. This means provider-sql could not reach database endpoints that require an HTTP CONNECT proxy, for example when running under Upbound Private Network Agent which injects proxy env vars to route traffic to private VPC endpoints.

## Implementation

Each driver uses its own dialer injection API:

- **PostgreSQL:** `pq.NewConnector` + `connector.Dialer(&httpProxyDialer{})` + `sql.OpenDB`
- **MySQL:** `gomysql.ParseDSN` + `cfg.DialFunc` + `gomysql.NewConnector` + `sql.OpenDB`
- **MSSQL:** `mssqldb.NewConnector` + `connector.Dialer` + `sql.OpenDB`

Proxy detection uses `http.ProxyFromEnvironment`, the same function the standard library uses, so `NO_PROXY` and other proxy env var semantics are respected automatically.

When no proxy is configured, `openDB` falls through to `sql.Open` with the same driver name and DSN as before. No other code runs. The stdlib itself uses this same pattern in `(*Transport).roundTrip`: check proxy URL, fall through to direct dial if nil.

The dialer injection APIs used here are each driver's documented extension point:

- **PostgreSQL:** `pq.Connector.Dialer` is the interface pq exports specifically for custom dialers
- **MySQL:** `mysql.Config.DialFunc` is the field go-sql-driver provides for dial interception
- **MSSQL:** `mssqldb.Connector.Dialer` is the field go-mssqldb exposes for custom dialers

Connections that do not match the proxy scope (via `NO_PROXY`, scheme filtering, etc.) are handled by `http.ProxyFromEnvironment` before any driver code is touched.

## Test plan

- [x] Unit tests pass: `go test ./pkg/clients/...`
- [x] No proxy configured: `openDB` falls through to `sql.Open`, behaviour identical to before (`TestOpenDBNoProxy`)
- [x] Proxy configured: tunnel completes HTTP CONNECT handshake and returns a usable connection (`TestTunnelSuccess`)
- [x] Proxy rejects connection: non-200 response surfaces as an error (`TestTunnelProxyRejected`


pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,62 @@
...
+
+type httpProxyDialer struct {
+	proxy *url.URL
+}
+
+func (d *httpProxyDialer) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	conn, err := (&net.Dialer{}).DialContext(ctx, "tcp", d.proxy.Host)
+	if err != nil {
+		return nil, err
+	}
```
fernandezcuesta, line 23:
```suggestion
	}
	defer func()
		if err != nil {
			closeErr := conn.Close()
			if closeErr != nil {
				err = fmt.Errorf("error closing connection (%w) after: %w", closeErr, err)
			}
		}
	}()
```

pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,62 @@
...
+	req := &http.Request{
+		Method: http.MethodConnect,
+		URL:    &url.URL{Host: addr},
+		Host:   addr,
+		Header: http.Header{},
+	}
+	if err := req.Write(conn); err != nil {
+		conn.Close() //nolint:errcheck
```
fernandezcuesta:
```suggestion
```

pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, err
+	}
+	resp, err := http.ReadResponse(bufio.NewReader(conn), req)
+	if err != nil {
+		conn.Close() //nolint:errcheck
```
fernandezcuesta: ### 
```suggestion
```

pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, err
+	}
+	resp.Body.Close() //nolint:errcheck
+	if resp.StatusCode != http.StatusOK {
+		conn.Close() //nolint:errcheck
```
fernandezcuesta:
```suggestion
```

fernandezcuesta (commented):
similarly for the rest

pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,69 @@
...
+		Method: http.MethodConnect,
+		URL:    &url.URL{Host: addr},
+		Host:   addr,
+		Header: http.Header{},
+	}
+	if err = req.Write(conn); err != nil {
+		return nil, err
+	}
+	var resp *http.Response
+	resp, err = http.ReadResponse(bufio.NewReader(conn), req)
```
chlunde: this will possibly hijack bytes from the database server - as NewReader does a 4 KB read and that could be some of the bytes not just from the proxy but the sql server banner data?

pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	resp.Body.Close() //nolint:errcheck
+	if resp.StatusCode != http.StatusOK {
+		err = fmt.Errorf("proxy CONNECT to %s: %s", addr, resp.Status)
+		return nil, err
+	}
+	return conn, nil
+}
+
+// openDB opens a *sql.DB for the given DSN. When HTTP_PROXY or HTTPS_PROXY is
```
chlunde, line 27: does it respect HTTP_PROXY? you construct a https URL below?

pkg/clients/mssql/proxy.go
```diff
@@ -0,0 +1,69 @@
...
+// set and applicable to the target endpoint, connections are tunnelled through
+// the proxy via HTTP CONNECT.
+func openDB(endpoint, port, dsn string) (*sql.DB, error) {
+	req, _ := http.NewRequest(http.MethodConnect, "https://"+endpoint+":"+port, nil)
```
chlunde: will this panic on invalid URL?

fernandezcuesta:
Also, shall we consolidate this to `pkg/clients/xsql` to avoid duplicated code?

pkg/clients/mssql/proxy.go
chlunde: it feels a bit weird to do this for users not setting up a proxy - but I see why your doing it

chlunde: You construct a &http.Request{ manually in the code the other place - should probably use the same here? this would avoid discarding the error too

Issue #203, opened 2024-11-28
Agreement on how to write missing connection parameters


pierluigilenoci:
There is a "small" issue that makes the UpJet AWS provider unusable in conjunction with the SQL provider. 
This is because there is a discrepancy between the way the UpJet provider writes credentials and how the SQL provider reads them.
Could the maintainers of both providers discuss and find a solution to unblock all users who would like to create databases and easily execute queries?
List of issues to help you analyze the problem:

- https://github.com/crossplane-contrib/provider-upjet-aws/issues/778
- https://github.com/crossplane-contrib/provider-upjet-aws/pull/798
- https://github.com/crossplane-contrib/provider-upjet-aws/issues/1189
- https://github.com/crossplane-contrib/provider-sql/pull/154
- https://github.com/crossplane-contrib/provider-sql/issues/191

pierluigilenoci:
The follow-up discussion can also be found here: https://crossplane.slack.com/archives/C01718T2476/p1732801643245679.

Duologic:
Thanks for brining this up @pierluigilenoci

_This was also discussed on the linked Slack thread but sharing here for posterity._

I've never seen this as a big issue as it was never a blocker for me, I've been using these through Compositions, which allow me to map `address` to `endpoint`.

Like this ([ref](https://github.com/crossplane-contrib/provider-sql/pull/154#issuecomment-1719578342)):

```
connectionDetails:
    - fromConnectionSecretKey: username
      name: username
    - fromConnectionSecretKey: attribute.password
      name: password
    - fromConnectionSecretKey: address
      name: endpoint
    - fromConnectionSecretKey: port
      name: port
```

I remembered vaguely that the provider for GCP supported the consistent endpoint/port/username/password but that doesn't seem to be the case for the Upjet provider. The secret from provider-upjet-gcp's CloudSQL connectionDetails does not match in any way that could possibly work with provider-sql without rewriting the secret (with a Composition for example).

The 'vanilla' provider-gcp did return the consistent secrets, hence where I remember that from, as well as the 'vanilla' provider-aws. You can see here that provider-aws maps `endpoint.address` to `endpoint`: https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/controller/rds/dbinstance/setup.go#L233 

So to be very pedantic, this was a regression introduced by provider-upjet-aws. :innocent:

[...]

PR #362, opened 2026-04-28
feat: add support for alternative connection secret formats

fernandezcuesta:
Add support to feed providerconfigs with custom format secrets.

Fixes: #203


chlunde (approved):
Cool! Could we add an example for documentation?

```yaml
  # Example: consume a connection secret that uses non-standard key names.
  #
  # provider-sql expects keys "endpoint", "port", "username" and "password" by
  # default. Secrets written by other providers (e.g. provider-upjet-aws RDS,
  # which writes "address", "port", "username", "password") can be consumed by
  # remapping the secret keys via spec.credentials.secretKeyMapping.
  #
  # Only specify the standard keys whose source key name differs from the
  # default — omitted entries fall through to the standard key.
  kind: ProviderConfig
  metadata:
    name: rds
  spec:
    sslMode: require
    credentials:
      source: PostgreSQLConnectionSecret
      connectionSecretRef:
        namespace: crossplane-system
        name: rds-connection
      secretKeyMapping:
        # Each field names the standard key; the value is the key to read from
        # the secret. Empty values fall through to the standard key.
        endpoint: address   # provider-upjet-aws writes the host as "address"
        # port:     port    # left as default
        # username: username
        # password: password
  ---
  # Connection secret in the shape produced by provider-upjet-aws RDS. The
  # fields above remap "address" to the standard "endpoint" key.
  apiVersion: v1
  kind: Secret
  metadata:
    name: rds-connection
    namespace: crossplane-system
  type: Opaque
  stringData:
    address: my-db.cluster-xyz.us-east-1.rds.amazonaws.com
    port: "5432"
    username: postgres
    password: hunter2
```

PR #363, opened 2026-04-30
feat(mysql/user): support non-default authentication plugins for IAM auth

### LINKED ISSUE #106: Support RDS IAM authentication

ronlevy1211:

Adds an optional `AuthenticationPlugin` field on the MySQL `User` resource. When set, the user is created with `CREATE USER ... IDENTIFIED WITH <plugin> [AS '<authString>']` instead of `IDENTIFIED BY <password>`. No password is generated, the connection secret omits the password key, and `ALTER USER ... IDENTIFIED BY` is skipped on Update so the provider does not downgrade plugin-auth users to native password auth.

Primary motivation is **AWS RDS IAM authentication**, where the DB user has no static password and clients fetch a short-lived auth token from AWS at connect time:

```yaml
spec:
  forProvider:
    authenticationPlugin:
      name: AWSAuthenticationPlugin
      authString: RDS
```

The same field also works for other auth plugins (`auth_socket`, `caching_sha2_password`, etc.).

#### Backward compatibility

When `AuthenticationPlugin` is unset, the existing password-based flow is preserved. Existing users / manifests are unaffected.

#### Notable changes

- New `AuthenticationPlugin{Name, AuthString}` struct on `UserParameters` (cluster + namespaced variants)
- `Create`: branches to `IDENTIFIED WITH` when the plugin is set; password generation skipped
- `Observe`: skips password drift check when the plugin is set (no password to drift)
- `UpdatePassword`: no-op when the plugin is set (avoids downgrading the user back to password auth)
- Plugin name is quoted as a SQL identifier; auth string is quoted as a SQL value
- Examples under `examples/cluster/mysql/user_iam_auth.yaml` and `examples/namespaced/mysql/user_iam_auth.yaml`
- Regenerated CRDs and deepcopy

Fixes #106

I have:

- Unit tests added for both cluster and namespaced reconcilers covering: create with plugin + authString, create with plugin only (no AS clause), update no-op for plugin-auth users
- End-to-end validated against a real AWS Aurora MySQL — full results in a comment below


ronlevy1211:
## E2E test results

Validated end-to-end against an AWS Aurora MySQL (Aurora 3 / MySQL 8.0 compatible) on a non-prod cluster. Built the provider from this branch as a custom OCI image and deployed it on Crossplane pointing at the cluster.

### Setup

```yaml
kind: User
metadata:
  name: iam-auth-test
  annotations:
    crossplane.io/external-name: iam-auth-test
spec:
  forProvider:
    authenticationPlugin:
      name: AWSAuthenticationPlugin
      authString: RDS
  providerConfigRef:
    name: mysql-shell-services
  writeConnectionSecretToRef:
    name: iam-auth-test-conn
    namespace: default
```

Plus a `Database` and `Grant` referring to the same user.

### Results

| Test | Result |
|---|---|
| User CR transitions to `Ready=True, Synced=True` | ✅ |
| `mysql.user` row shows `plugin=AWSAuthenticationPlugin` (proves the SQL we emit is accepted by AWS RDS) | ✅ |
| Connection secret has no password key | ✅ |
| `Grant` and `Database` reconcile successfully alongside the IAM-auth user | ✅ |
| No drift over 3 consecutive reconcile cycles (90s) — `metadata.resourceVersion` unchanged, no `ALTER USER` in provider logs (validates the `Observe` and `UpdatePassword` skip-when-plugin paths) | ✅ |
| Deleting the User CR runs `DROP USER` — `SELECT … FROM mysql.user WHERE user = 'iam-auth-test'` returns no rows | ✅ |

Direct DB query proving the plugin was applied:

```
mysql> SELECT user, host, plugin FROM mysql.user WHERE user = 'iam-auth-test';
+---------------+------+--------------------------+
| user          | host | plugin                   |
+---------------+------+--------------------------+
| iam-auth-test | %    | AWSAuthenticationPlugin  |
+---------------+------+--------------------------+
```

ronlevy1211:
Hi @fernandezcuesta @chlunde @JevgeniF — friendly ping for a first pass when you have a moment. The PR adds `AuthenticationPlugin` to the MySQL `User` resource so the user can be created with non-default plugins (specifically the AWS RDS `AWSAuthenticationPlugin/RDS` for IAM auth). E2E results against Aurora MySQL 8.0 are documented in the comment above.

apis/namespaced/mysql/v1alpha1/user_types.go
```diff
@@ -38,10 +38,29 @@ type UserStatus struct {
 // UserParameters define the desired state of a MySQL user instance.
 type UserParameters struct {
 	// PasswordSecretRef references the secret that contains the password used
-	// for this user. If no reference is given, a password will be auto-generated.
+	// for this user. If no reference is given and AuthenticationPlugin is
+	// unset, a password will be auto-generated.
+	// Ignored when AuthenticationPlugin is set.
```
fernandezcuesta: Can those be explicitly set mutually exclusive?
I mean either kubebuilder annotation or lateinitializer ignoredFields.
If you're able to restrict that from the API/CRD, the logic can be simplified a bit

ronlevy1211: Done — pushed af1d37f. Added a `+kubebuilder:validation:XValidation` rule on `UserParameters` that rejects setting both `passwordSecretRef` and `authenticationPlugin` (matches the pattern used by MSSQL `User`). The runtime branches in `Observe`/`Create`/`UpdatePassword` stay because they encode the semantic that plugin-auth users have no password, not a defensive guard. Doc comments on both fields tightened to drop the "ignored when..." wording.

ronlevy1211:
Hey @fernandezcuesta — no rush, just a gentle nudge in case this slipped past. Pushed af1d37f last Monday with the `+kubebuilder:validation:XValidation` rule per your suggestion — cleaned up the implementation nicely. CI is green and the change is small. Whenever you have a few minutes for a re-look, much appreciated 🙏

Also pinging @chlunde @JevgeniF for any second opinion if Fernando is heads-down elsewhere.

pkg/controller/cluster/mysql/user/reconciler.go
chlunde, line 215: Could you get `plugin` here for drift detection and add support for changing method in `Update`? crossplane managed resources should be able to update any field as long as the upstream system supports it, so we don't silently ignore some types of changes

chlunde (changes requested):
Could you add drift detection, or did you already discuss this?

pkg/controller/cluster/mysql/user/reconciler.go
ronlevy1211: Done - pushed 12cd322. Three pieces:

1. **Observe** now SELECTs `plugin, authentication_string` alongside the resource-options columns. Hydrates `observed.AuthenticationPlugin` only when the observed plugin is non-default-password (`caching_sha2_password` / `mysql_native_password` / `sha256_password` are treated as "user has a password" and left as nil - they're the implicit default and would otherwise force drift on every reconcile of a password user).

2. **`upToDate`** compares `AuthenticationPlugin` (name + authString, nil-aware) ahead of the resource-options checks so plugin drift is caught even when `ResourceOptions` is nil.

3. **Update** detects plugin drift via `Status.AtProvider.AuthenticationPlugin` (cached by Observe) and handles three transitions:
   - **password → plugin**: emits `ALTER USER ... IDENTIFIED WITH <plugin> [AS '<authString>']` (new `executeAlterUserWithPluginQuery` helper, mirrors the Create-time quoting - identifier-quote plugin name, value-quote authString).
   - **plugin → password**: falls through to the existing `UpdatePassword` path. `ALTER USER ... IDENTIFIED BY '<pw>'` both sets the password and restores the default password plugin in one statement.
   - **same plugin, different authString**: same `ALTER USER ... IDENTIFIED WITH` query with the new auth string.

The API-level `XValidation` rule already enforces mutual exclusivity of `passwordSecretRef` and `authenticationPlugin`, so the spec always represents one mode and the transition is unambiguous.

Tests cover the three transitions above plus an updated `AuthenticationPluginSkipsAlterPassword` that now sets observed == desired and asserts neither `IDENTIFIED BY` nor `IDENTIFIED WITH` runs.

ronlevy1211:
Hi @chlunde, gentle bump on this. Drift detection from your last review is added (commit `12cd3225`), and all CI checks are green. Noticed v0.16.0-beta.0 was cut yesterday, so it would be great to land this before the stable release if timing works. Happy to address anything else.

ronlevy1211:
Hi @chlunde, picking this up after some time away. To recap what changed since your last review: drift detection was added in `12cd3225` — `Observe()` now reads back `mysql.user.plugin` (and `authentication_string` where applicable) and reports `ResourceUpToDate: false` when either drifts from spec. `Update()` rewrites both via `ALTER USER ... IDENTIFIED WITH ...`. Test coverage is in the same commit; the follow-up `a3aedb1` keeps `Update()` under the gocyclo limit and regenerates the CRDs. All CI green.

Could you let me know if this matches what you had in mind, or if there's a different shape you'd prefer? Happy to iterate.

pkg/controller/cluster/mysql/user/reconciler.go
```diff
@@ -245,10 +272,64 @@ func (c *external) Observe(ctx context.Context, mg *v1alpha1.User) (managed.Exte
...
 
+// isPasswordPlugin reports whether the given MySQL plugin name is one of the
+// password-based authentication plugins. Users created with IDENTIFIED BY
+// '<pwd>' are assigned a password plugin (caching_sha2_password is the MySQL
+// 8.0 default; mysql_native_password and sha256_password are common). For
+// these we treat AuthenticationPlugin as nil in the observed state because
+// they represent "use a password" rather than an opt-in plugin choice.
+func isPasswordPlugin(name string) bool {
+	switch name {
+	case "caching_sha2_password", "mysql_native_password", "sha256_password":
```
fernandezcuesta, line 286: Not that I like hardcoded plugin names but can't find another solution

fernandezcuesta (approved):
LGTM

pkg/controller/cluster/mysql/user/reconciler.go
chlunde: Thanks!

chlunde:
Claude found another issue, can you check it? Setting spec.forProvider.authenticationPlugin.name to a password plugin (mysql_native_password, caching_sha2_password, sha256_password) makes the MySQL User reconcile forever, because Observe never hydrates status.atProvider.authenticationPlugin for those plugins, so upToDate can never return true.

ronlevy1211:
Good catch, thanks - confirmed. `Observe` was unconditionally treating password plugins as "no plugin", so when the spec explicitly requested one (`mysql_native_password` etc.) the observed value stayed nil and `upToDate` could never converge, re-issuing `ALTER USER` on every reconcile.

Fixed in a686fbe: the observed plugin is now hydrated whenever the spec opts into a plugin, so it converges for password and non-password plugins alike; the "no plugin requested" path is unchanged. Added a regression test that reproduces the infinite reconcile, and applied the same fix to the namespaced reconciler since it had the identical logic. All green locally (build/vet/tests).

chlunde (approved):

fernandezcuesta:
Thanks @ronlevy1211 for your contribution! :rocket:

PR #345, opened 2026-03-18
feat(pgsql): extend Grant kind specification with schema, tables, columns, sequences, routines, foreign data wrappers, foreign servers

fernandezcuesta:
### Description of your changes

This is a follow up of #235 with some minor issues found while merging from master.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

Added tests

[contribution process]: https://git.io/fj2m9

Issue #162, opened 2023-11-22
specific list of mysql privileges getting revoked during reconciliation


luis-alen:
### What happened?

When provider-sql for mysql grants the below specific list of privileges to a user, it revokes all of them and then grants again when reconciling although the grant object has not been changed at all. This causes the user to temporarily lose access to the database.

```
      - SELECT
      - INSERT
      - UPDATE
      - DELETE
      - CREATE
      - DROP
      - REFERENCES
      - INDEX
      - ALTER
      - CREATE TEMPORARY TABLES
      - LOCK TABLES
      - EXECUTE
      - CREATE VIEW
      - SHOW VIEW
      - CREATE ROUTINE
      - ALTER ROUTINE
      - EVENT
      - TRIGGER
 ```

If you grep for the username on mysql general logs, you'll see the following:

```
2023-11-22T10:51:40.534820Z	   68 Query	SHOW GRANTS FOR 'my-app'@'%'
2023-11-22T10:51:40.542055Z	   69 Query	REVOKE ALL ON `my-app`.* FROM 'my-app'@'%'
2023-11-22T10:51:40.586787Z	   71 Query	GRANT ALTER, ALTER ROUTINE, CREATE, CREATE ROUTINE, CREATE TEMPORARY TABLES, CREATE VIEW, DELETE, DROP, EVENT, EXECUTE, INDEX, INSERT, LOCK TABLES, REFERENCES, SELECT, SHOW VIEW, TRIGGER, UPDATE ON `my-app`.* TO 'my-app'@'%'
```

However, if you grant only `SELECT` and `INSERT`  or `ALL PRIVILEGES`, for example, this won't happen. In that case crossplane seems to understand there's no diff and doesn't take any action, as expected.

I'd say this is an extension of https://github.com/crossplane-contrib/provider-sql/issues/126, which was fixed by https://github.com/crossplane-contrib/provider-sql/pull/136, but apparently not fully.

### How can we reproduce it?

- enable mysql general log
- create a database

```
apiVersion: mysql.sql.crossplane.io/v1alpha1
kind: Database
metadata:
  name: my-app
spec:
  providerConfigRef:
    name: mysql8
```

- create a user

```
apiVersion: mysql.sql.crossplane.io/v1alpha1
kind: User
metadata:
  name: my-app
spec:
  providerConfigRef:
    name: mysql8
  deletionPolicy: Delete
  forProvider: {}
  writeConnectionSecretToRef:
[...]

PR #376, opened 2026-05-18
374  support server versions xsql

### LINKED ISSUE #374: Support server version detection for MSSQL+MySQL
In [749c80a](https://github.com/crossplane-contrib/provider-sql/commit/749c80a) we added the ability to filter privileges based on PostgreSQL server version (e.g., the `MAINTAIN` privilege is only available in PG 17+).

The reconcilers call `GetServerVersion()` during `Connect()` and pass the version to privilege expansion functions.

Both MySQL and MSSQL had `GetServerVersion` added to satisfy the `xsql.DB` interface, but the implementations were no-ops.

Real `GetServerVersion` implementations to allow a similar experience for MSSQL and MySQL.

fernandezcuesta:

Add support for running tests with custom versions for mariadb and mssql.

Will not wire it up yet (as we have for Pg, which checks the version on every connect) until this is actually used.

Fixes: #374

I have:

n/a, same logic as for Postgres.


pkg/clients/xsql/common_test.go
```diff
@@ -110,3 +112,44 @@ func TestRemapCredentialKeys(t *testing.T) {
...
+		{name: "MySQL_9", version: "9.1.0", want: 90100},
+		{name: "MariaDB", version: "10.11.6-MariaDB", want: 101106},
+
+		// MSSQL-style versions (build number > 99, ignored)
+		{name: "MSSQL_2022", version: "16.0.1125.1", want: 160000},
+		{name: "MSSQL_2019", version: "15.0.4355.3", want: 150000},
+		{name: "MSSQL_NonZeroMinor", version: "16.5.100.1", want: 160500},
+
+		// PostgreSQL-style (for reference, PG uses its own query)
+		{name: "ThreeDigitPatch", version: "14.2.1", want: 140201},
```
chlunde: `"14.2.1" → 140201` — I think PG would report `140002` here? `server_version_num` is `major*10000 + minor` since PG 10 (on postgres:18 it's `180006`). So the `160200` in the postgres client's comment is wrong and has been for a while, and this matches the comment rather than the server :thinking:

pkg/clients/xsql/common.go
```diff
@@ -25,6 +28,41 @@ type DB interface {
 	GetServerVersion(ctx context.Context) (int, error)
 }
 
+// ParseVersion parses a database version string into an integer
+// encoded as major*10000 + minor*100 + patch.
+// Suffixes after '-' are stripped (e.g. "8.0.35-ubuntu" → 80035).
+// Patch values above 99 are ignored (e.g. MSSQL build numbers).
+func ParseVersion(version string) (int, error) {
+	if idx := strings.IndexByte(version, '-'); idx >= 0 {
```
chlunde: For #162 I think we need to tell MariaDB from MySQL, not just how new it is — their database-level `ALL` differs, MariaDB 12 also has `DELETE HISTORY` and `SHOW CREATE ROUTINE`. This cuts at the first `-`, which is exactly where `MariaDB` sits. Could we keep the raw string around too?

cluster/local/mssql.server.yaml
```diff
@@ -28,7 +28,7 @@ spec:
     spec:
       containers:
       - name: mssql
-        image: mcr.microsoft.com/mssql/server:2019-CU32-ubuntu-20.04
+        image: mcr.microsoft.com/mssql/server:${MSSQL_VERSION}
```
chlunde: I don't think `MSSQL_VERSION` reaches here — `setup_mssql` seds the whole image line and puts `${MSSQL_IMAGE}` in it, which still defaults to 2019-CU32. Could we reuse `MSSQL_IMAGE` rather than add a second variable?

pkg/controller/namespaced/mysql/grant/reconciler.go
```diff
@@ -121,6 +121,8 @@ func (c *connector) Connect(ctx context.Context, mg *namespacedv1alpha1.Grant) (
 		return nil, errors.Wrap(err, errTLSConfig)
 	}
 
+	// To add version-gated logic, call db.GetServerVersion(ctx) here
```
chlunde: This is a second `sql.Open` per reconcile if anyone follows it — a mysql grant reconcile with no drift is one query today, so it would double. Same concern I had on #345, except now it'd be in twelve places. Worth caching it per DSN, or waiting for one of the pooling PRs, before we recommend it? (also, no `db` in scope here, it's built inline in the return)

.github/workflows/ci.yml
```diff
@@ -245,7 +258,11 @@ jobs:
         #  BUILD_ARGS: "--load"
 
       - name: Run E2E Tests
-        run: make e2e USE_HELM=true POSTGRES_VERSION=${{ inputs.postgres_version || '18' }}
+        run: >-
+          make e2e USE_HELM=true
+          POSTGRES_VERSION=${{ inputs.postgres_version }}
```
chlunde: Dropping the `|| '18'` makes this `POSTGRES_VERSION=` on push and PR, and an empty value on the make line beats the `?=` in the Makefile. It still works because the shell functions default again, but the Makefile defaults are then dead. Could the `concurrency:` block and the quoting cleanup go in #454 instead? I'd like this one small since #162 is waiting on it.

pkg/clients/xsql/common_test.go
```diff
@@ -110,3 +112,44 @@ func TestRemapCredentialKeys(t *testing.T) {
...
+		{name: "PatchExactly99", version: "14.0.99", want: 140099},
+		{name: "PatchOver99", version: "14.0.100", want: 140000},
+
+		// Errors
+		{name: "InvalidFormat", version: "invalid", wantErr: true},
```
chlunde: worth a `{version: "8", wantErr: true}` here too — both cases below error via `strconv` anyway, so nothing actually exercises the `len(parts) < 2` guard: relaxing it to `< 1` still passes the whole table, and it panics on `"8"`.

go.mod
```diff
@@ -23,6 +23,8 @@ require (
 	sigs.k8s.io/controller-tools v0.22.0
 )
 
+require github.com/stretchr/testify v1.12.1
```
chlunde: we use `go-cmp` in about 25 test files — is testify worth a new direct dependency for one table?

pkg/clients/xsql/common_test.go
chlunde: `"14.2.1" → 140201` — I think PG would report `140002` here? `server_version_num` is `major*10000 + minor` since PG 10 (on postgres:18 it's `180006`). So the `160200` in the postgres client's comment is wrong and has been for a while, and this matches the comment rather than the server :thinking:

pkg/clients/xsql/common.go
chlunde: For #162 I think we need to tell MariaDB from MySQL, not just how new it is — their database-level `ALL` differs, MariaDB 12 also has `DELETE HISTORY` and `SHOW CREATE ROUTINE`. This cuts at the first `-`, which is exactly where `MariaDB` sits. Could we keep the raw string around too?

cluster/local/mssql.server.yaml
chlunde: I don't think `MSSQL_VERSION` reaches anything — `setup_mssql` still seds `image: mcr.microsoft.com/mssql/server:.*` and replaces it with `${MSSQL_IMAGE}`, so it overwrites the placeholder you added here. Could we pick one of them? envsubst rather than sed perhaps, since that's what postgres and mariadb already do.

One thing to watch either way: `MSSQL_IMAGE` defaults to 2019-CU32 and the new `MSSQL_VERSION` to 2022-CU24, so making this work also changes which SQL Server CI actually tests — probably good, but worth doing on purpose.

pkg/controller/namespaced/mysql/grant/reconciler.go
chlunde: This is a second `sql.Open` per reconcile if anyone follows it — a mysql grant reconcile with no drift is one query today, so it would double. Same concern I had on #345, except now it'd be in twelve places. Worth caching it per DSN, or waiting for one of the pooling PRs, before we recommend it? (also, no `db` in scope here, it's built inline in the return)

.github/workflows/ci.yml
chlunde: Dropping the `|| '18'` makes this `POSTGRES_VERSION=` on push and PR, and an empty value on the make line beats the `?=` in the Makefile. It still works because the shell functions default again, but the Makefile defaults are then dead. Could the `concurrency:` block and the quoting cleanup go in #454 instead? I'd like this one small since #162 is waiting on it.

pkg/clients/xsql/common_test.go
chlunde: worth a `{version: "8", wantErr: true}` here too — both cases below error via `strconv` anyway, so nothing actually exercises the `len(parts) < 2` guard: relaxing it to `< 1` still passes the whole table, and it panics on `"8"`.

go.mod
chlunde: we use `go-cmp` in about 25 test files — is testify worth a new direct dependency for one table?

pkg/clients/xsql/common_test.go
chlunde, line 134: `"14.2.1" → 140201` — I think PG would report `140002` here? `server_version_num` is `major*10000 + minor` since PG 10:

```
$ docker run -d --rm --name pg -e POSTGRES_PASSWORD=x postgres:14.2
$ docker exec pg psql -U postgres -Atc \
    "SELECT current_setting('server_version'), current_setting('server_version_num')"
14.2 (Debian 14.2-1.pgdg110+1)|140002

# and on postgres:18
18.6 (Debian 18.6-1.pgdg13+2)|180006
```

So the `160200` in the postgres client's doc comment is wrong and has been for a while, and this matches the comment rather than the server 🤔

pkg/clients/xsql/common.go
chlunde, line 36: For #162 I think we need to tell MariaDB from MySQL, not just how new it is — their database-level `ALL` differs:

```
$ docker run -d --rm --name my -e MYSQL_ROOT_PASSWORD=x mysql:8.4
$ docker exec my mysql -uroot -px -e \
    "CREATE DATABASE d3; CREATE USER u3@'%' IDENTIFIED BY 'p';
     GRANT ALL ON d3.* TO u3@'%'; REVOKE TRIGGER ON d3.* FROM u3@'%';
     SHOW GRANTS FOR u3@'%';"
8.4.11
GRANT SELECT, ..., CREATE ROUTINE, ALTER ROUTINE, EVENT ON `d3`.* TO `u3`@`%`

# mariadb:12 — same statements
12.3.3-MariaDB-ubu2404
GRANT SELECT, ..., CREATE ROUTINE, ALTER ROUTINE, EVENT,
      DELETE HISTORY, SHOW CREATE ROUTINE ON `d3`.* TO `u3`@`%`
```

MariaDB 12 has `DELETE HISTORY` and `SHOW CREATE ROUTINE` on top. This cuts at the first `-`, which is exactly where `MariaDB` sits in that version string. Could we keep the raw string around too?

cluster/local/mssql.server.yaml
chlunde, line 31: I don't think `MSSQL_VERSION` reaches anything — `setup_mssql` still seds the image line and replaces it with `${MSSQL_IMAGE}`. Could we pick one? envsubst like postgres/mariadb perhaps, but it'd need a `${MSSQL_VERSION:-2019-CU32-ubuntu-20.04}` default the way they have one — CI passes it empty on push/PR (see the ci.yml comment), so otherwise the tag ends up blank.

pkg/controller/namespaced/mysql/grant/reconciler.go
chlunde: This is a second `sql.Open` per reconcile if anyone follows it — a mysql grant reconcile with no drift is one query today, so it would double. Same concern I had on #345, except now it'd be in twelve places. Worth caching it per DSN, or waiting for one of the pooling PRs, before we recommend it? (also, no `db` in scope here, it's built inline in the return)

.github/workflows/ci.yml
chlunde: Dropping the `|| '18'` makes this `POSTGRES_VERSION=` on push and PR events, and an empty value on the make line beats the `?=` in the Makefile. It still works because the shell functions default again, but the Makefile defaults are then dead.

pkg/clients/xsql/common_test.go
chlunde, line 140: worth a `{version: "8", wantErr: true}` here too — both cases below error via `strconv` anyway, so nothing actually exercises the `len(parts) < 2` guard: relaxing it to `< 1` still passes the whole table, and it panics on `"8"`.

go.mod
chlunde: we use `go-cmp` in about 25 test files — is testify worth a new direct dependency for one table?

chlunde (changes requested):

pkg/clients/xsql/common_test.go
fernandezcuesta: yeah good catch

pkg/clients/xsql/common.go
fernandezcuesta: I guess something we can do is instead of returning the version as a number, return a struct.

I'd prefer to keep this PR focused on the version plumbing and defer the MariaDB detection to #162, where we'll actually need it.

When that lands, the plan would be to change `GetServerVersion` to return a VersionInfo struct:

```go
type VersionInfo struct {
    Number int    // e.g. 101106
    Raw    string // e.g. "10.11.6-MariaDB"
    Engine string // "mysql", "mariadb", "postgresql", "mssql"
  }
```

That way `ParseVersion` stays a pure number parser, and engine detection lives in each client's `GetServerVersion` where it belongs. The interface change touches
all engines, so it makes more sense as part of the PR that actually consumes it, or in a separate one due to the large number of files affected.

Does that work, or would you rather we implement this here?

pkg/controller/namespaced/mysql/grant/reconciler.go
fernandezcuesta: removed the comment from all 12 reconcilers, but the only option we've is a kind of per-DSN caching, but that comes with a cost too. Shall we track this in a separate issue?

pkg/clients/xsql/common_test.go
fernandezcuesta: done

PR #239, opened 2025-05-30
Use native Golang connection pooling to reduce number of connections opened for postgres (#195)

ajcaldera1:
### Description of your changes
This change reduces the number of connections that must be opened to support a reconciliation job. On execution of New() it opens a connection pool with a maximum of 5 connections (hard-coded in this iteration) with a max lifetime of 10 minutes. After 2 minutes, any idle (unused) connections will be terminated to a minimum of 2 connections. These connections are then reused across the entire lifecycle. 

Fixes #195 

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

e2e tests done and tested by production observations over the last 7 months. 

[contribution process]: https://git.io/fj2m9

Issue #220, opened 2025-03-04
MySQL: Databases not synced because of max_prepared_stmt_count


jscheytt:
### What happened?

We are using MySQL Databases, Users, and Grants resources to manage them in a Percona XtraDB operated MySQL server. (Nothing fancy, around 60 resources for each of the 3 types.)

All non-prod environments work fine, but in our production environment, the all Databases, Users, and Grants become ready (as indicated by the `READY=True` column) but Databases and Users do not become synced (as indicated by the `SYNCED=False` column). The events of the resources show e.g. this error:

```
cannot select database: Error 1461: Can't create more than max_prepared_stmt_count statements (current value: 16382)
```

We have not tried increasing this value yet because we found out the following:
1. This issue in go-sql-driver/mysql indicates that others have run into this problem before as well: https://github.com/go-sql-driver/mysql/issues/701
2. Their maintainers' recommendation (I simplify to what I have grasped) is that you should definitely set these "Important settings": https://github.com/go-sql-driver/mysql?tab=readme-ov-file#important-settings
3. I don't see these anywhere in the code. I suppose they should probably happen here (in all calls of `sql.Open()`? https://github.com/crossplane-contrib/provider-sql/blob/33fb91cad4fe5172224bd38e7ace3c5e30a12e2c/pkg/clients/mysql/mysql.go#L78C12-L78C20

Can we work towards implementing these "important settings"? I am happy to contribute if possible.

### How can we reproduce it?

Sorry, I don't have very concrete steps for reproduction.
Probably try a stressed MySQL server with lots of data and connections?

### What environment did it happen in?
* Crossplane version: v1.19.2
* Crossplane SQL Provider version: v0.11.0
* Cloud provider: Hetzner
* Kubernetes version: v1.30.6+k3s1

PR #381, opened 2026-05-27
feat(mysql): configurable connection pool + DSN dial timeout

erikmiller-gusto:
## Problem

`provider-sql` v0.14.0 has no MySQL connection pool configuration. The MySQL client calls `sql.Open` + `db.Close` on every `Exec` / `Query` / `Scan`, which means a fresh TCP+TLS+MySQL handshake per call. Combined with the default `--max-reconcile-rate=10` per controller × 3 controllers (Database/User/Grant), in-flight handshake count climbs into the dozens during bursts. Downstream proxies — AWS RDS Proxy under load, smaller RDS instances, Percona XtraDB clusters, LocalStack's MySQL proxy — can't service that storm and start aborting handshakes or hitting `max_prepared_stmt_count`. Once the pool lands in an `invalid connection` state, the lack of pooling means every subsequent reconcile hits the same broken pool until backoff eventually clears.

This is the failure shape described in:

- [#110](https://github.com/crossplane-contrib/provider-sql/issues/110) — "Avoid reconciling many objects on the same database at the same time" — quotes the Go docs directly:
  > The returned DB is safe for concurrent use by multiple goroutines and maintains its own pool of idle connections. Thus, the Open function should be called just once.
- [#195](https://github.com/crossplane-contrib/provider-sql/issues/195) — "Provider should use connection pooling or single transaction to minimize load on database when reconciling"
- [#220](https://github.com/crossplane-contrib/provider-sql/issues/220) — "MySQL: Databases not synced because of `max_prepared_stmt_count`" — references the [go-sql-driver "Important settings"](https://github.com/go-sql-driver/mysql?tab=readme-ov-file#important-settings) that this PR implements.

The current code defeats Go's "Open should be called just once" guarantee.

## What this PR changes

1. **Process-wide pool cache** (`pkg/clients/mysql/mysql.go`) — `map[dsn]*sql.DB` shared across reconciles. `mysql.New` / `NewWithConfig` look up by DSN and reuse the cached `*sql.DB`. Credential rotation lands a fresh pool naturally because DSN changes.

2. **Per-pool tuning** (`NewWithConfig` + `ConnectionPoolConfig`) — optional knobs for `MaxOpenConns`, `MaxIdleConns`, `ConnMaxLifetime`, `ConnMaxIdleTime`, `DialTimeout`. Nil config preserves Go defaults — opt-in, no behavior change for existing users.

3. **DSN dial timeout** (`DSNWithDialTimeout`) — when `DialTimeout > 0`, appends `&timeout=Xs` to the DSN, which go-sql-driver maps to the `net.Dialer.Timeout`. Previously the driver had no dial timeout, so a hung TCP connect blocked the reconciler indefinitely.

4. **ProviderConfig API exposure** (`apis/cluster`, `apis/namespaced`) — new optional `connectionPool` field on MySQL `ProviderConfigSpec` / `ClusterProviderConfigSpec` with the same five tunables. CRDs regenerated.

5. **Reconciler plumbing** — all 6 MySQL reconcilers (Database/User/Grant × cluster/namespaced) read `pc.Spec.ConnectionPool`, convert via `ConnectionPoolSpec.ToPoolValues` + `mysql.NewConnectionPoolConfig`, and pass through to `NewWithConfig`. `ProviderInfo` carries the spec for the namespaced reconcilers.

6. **Driver blank-import** — added `_ \"github.com/go-sql-driver/mysql\"` to `mysql.go` so the package's `sql.Open` calls have the driver registered without relying on a transitive import from `reconciler.go` or `tls.go`. Existing code was depending on import-graph coincidence.

## Backward compatibility

- `mysql.New(creds, tls, binlog)` signature unchanged — wraps `NewWithConfig(creds, tls, binlog, nil)`.
- `DSN(...)` signature unchanged; new `DSNWithDialTimeout(...)` is additive.
- ProviderConfig additions are all optional. A ProviderConfig without `connectionPool` produces the same client behavior as before, **modulo the pool cache itself** — which is the intended fix per #110.
- The pool cache changes one observable thing: connections persist between reconciles as idle entries (per Go's defaults: up to 2 per pool). DB admins watching connection counts will see the per-reconcile create/destroy pattern replaced by a small steady-state pool. This is the documented correct behavior.

## Example usage

```yaml
kind: ProviderConfig
metadata:
  name: my-mysql
spec:
  credentials:
    source: MySQLConnectionSecret
    connectionSecretRef:
      name: my-mysql-credentials
      namespace: crossplane-system
  connectionPool:
    maxOpenConns: 10
    maxIdleConns: 5
    connMaxLifetime: 5m
    connMaxIdleTime: 1m
    dialTimeout: 10s
```

## Tests

- 10 new unit tests in `pkg/clients/mysql/mysql_test.go`:
  - DSN dial-timeout formatting (zero / positive / negative / coexists-with-binlog)
  - Pool settings applied via `db.Stats()`
  - Nil config uses Go defaults
  - Pool cache reuse for identical DSNs (the core architectural fix)
  - Distinct pools for distinct DSNs
  - Backward-compatible `New()` behavior
- All 6 reconciler test packages updated for the `newDB` field signature change (function-type-only update; no test logic changes).
- `go test ./...`: **690 passed** (was 680 baseline → +10 new).
- `go vet ./...`: clean.
- Local `golangci-lint` skipped — version skew with the Makefile-pinned 2.10.1 — CI will run the correct version.

## Out of scope

- Postgres + MSSQL clients have the same architectural gap. Deliberately MySQL-only here to keep the diff reviewable. Issue [#239](https://github.com/crossplane-contrib/provider-sql/issues/239) was a stale Postgres-only attempt at pooling; the pattern in this PR is straightforward to port if maintainers want a separate PR for it.
- No changes to reconciler logic — pure SQL-driver-layer fix.
- No CLI flag exposure; configuration is per-ProviderConfig, which fits per-tenant tuning better than a single process-wide flag.


fernandezcuesta:
I think the pool logic will be replaced by #434, but we can keep this PR (or create a new one) for the dial timeout support

Issue #131, opened 2023-03-01
GCPCloudSQL Role Password unable to update


gogovan-vincentngai:
### What happened?
GCPCloudSQL Role Password unable to update

### How can we reproduce it?
- GCP Create CloudSQL Instance
- Crate DB inside the instance
- Create User Role with below config 
```
apiVersion: postgresql.sql.crossplane.io/v1alpha1
kind: Role
metadata:
  annotations:
    crossplane.io/external-name: vincent-test
  name: vincent-test
  namespace: infra-tools
spec:
  forProvider:
    passwordSecretRef:
      name: vincent-test
      namespace: infra-tools
      key: password
    connectionLimit: -1
    privileges:
      login: true
  providerConfigRef:
    name: provider-config
```
- secret  # aaaabbbbbb123123
```
apiVersion: v1
data:
  password: YWFhYWJiYmJiYjEyMzEyMw==  
kind: Secret
metadata:
  labels:
    argocd.argoproj.io/instance: crossplane-cloudsql
  name: vincent-test
  namespace: infra-tools
type: connection.crossplane.io/v1alpha1
```
- Test user `success login` with the password 
- Update the secret to otherthings
- Wait it sync  , event message will show `UpdatedExternalResource` something updated 
- Retest user with new password 

### What environment did it happen in?
Crossplane version: 1.10.1
provider-sql: 0.6.0
Kubernetes version: 1.22
Kubernetes distribution: GKE

pascal-sochacki:
Hey, I just looked at the bug and found you need to define a `writeConnectionSecretToRef` in your Role

example:
```
apiVersion: postgresql.sql.crossplane.io/v1alpha1
kind: Role
metadata:
  annotations:
    crossplane.io/external-name: vincent-test
  name: vincent-test
spec:

  writeConnectionSecretToRef: <<<
    name: db-pass-2                   <<<
    namespace: default               <<<

  forProvider:
    passwordSecretRef:
      name: db-pass
      namespace: default
      key: password
    connectionLimit: -1
    privileges:
      login: true
```

The Reason for this is, in the current implementation, the difference is detected if the password in the created secret (from the  `writeConnectionSecretToRef`) is not the same as in the `passwordSecretRef `. (see the code at https://github.com/crossplane-contrib/provider-sql/blob/c6c358b9f05e1e7f2e77906c20eb5d714e29143e/pkg/controller/postgresql/role/utils.go#L47). I hope I could help.

PR #394, opened 2026-06-04
feat: Trigger PostgreSQL Role reconciliation when password Secret changes

Mirt0x:

- Added Secret watches to Role controllers so Roles referencing a password Secret are enqueued when that Secret changes. See the namespaced and cluster Role controllers.
- Extracted the mapping logic into helpers `namespacedSecretMapper` and `clusterSecretMapper` that list Roles and return reconcile requests for Roles whose `PasswordSecretRef` matches the Secret.
- Added unit tests for the mapper behavior (including tests that exercise Secret-data changes).
- Ran formatting on modified files.

I have:

- `go test ./pkg/controller/namespaced/postgresql/role ./pkg/controller/cluster/postgresql/role`
both packages passed locally
- `make reviewable`

](https://git.io/fj2m9)

### DIFF
--- a/pkg/controller/cluster/postgresql/role/reconciler.go
+++ b/pkg/controller/cluster/postgresql/role/reconciler.go
@@ -32,6 +32,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/v2/apis/common/v1"
 	xpcontroller "github.com/crossplane/crossplane-runtime/v2/pkg/controller"
@@ -65,6 +67,20 @@ const (
 	maxConcurrency = 5
 )
 
+func clusterSecretMapper(kube client.Client, s *corev1.Secret) []reconcile.Request {
+	var rl v1alpha1.RoleList
+	if err := kube.List(context.Background(), &rl); err != nil {
+		return nil
+	}
+	reqs := make([]reconcile.Request, 0)
+	for _, role := range rl.Items {
+		if role.Spec.ForProvider.PasswordSecretRef != nil && role.Spec.ForProvider.PasswordSecretRef.Name == s.Name && role.Spec.ForProvider.PasswordSecretRef.Namespace == s.Namespace {
+			reqs = append(reqs, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: role.Namespace, Name: role.Name}})
+		}
+	}
+	return reqs
+}
+
 // Setup adds a controller that reconciles Role managed resources.
 func Setup(mgr ctrl.Manager, o xpcontroller.Options) error {
 	name := managed.ControllerName(v1alpha1.RoleGroupKind)
@@ -89,13 +105,24 @@ func Setup(mgr ctrl.Manager, o xpcontroller.Options) error {
 	)); err != nil {
 		return err
 	}
-	return ctrl.NewControllerManagedBy(mgr).
+	builder := ctrl.NewControllerManagedBy(mgr).
 		Named(name).
 		For(&v1alpha1.Role{}).
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: maxConcurrency,
-		}).
-		Complete(r)
+		})
+
+	// Watch Secrets and enqueue cluster Roles that reference the secret as
+	// their PasswordSecretRef so password Secret changes trigger reconciliation.
+	builder = builder.Watches(&corev1.Secret{}, handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
+		s, ok := obj.(*corev1.Secret)
+		if !ok {
+			return nil
+		}
+		return clusterSecretMapper(mgr.GetClient(), s)
+	}))
+
+	return builder.Complete(r)
 }
 
 type connector struct {
--- a/pkg/controller/cluster/postgresql/role/reconciler_test.go
+++ b/pkg/controller/cluster/postgresql/role/reconciler_test.go
@@ -80,6 +80,70 @@ func (m mockDB) GetServerVersion(ctx context.Context) (int, error) {
 	return m.MockGetServerVersion(ctx)
 }
 
+func Test_clusterSecretMapper(t *testing.T) {
+	secret := &corev1.Secret{}
+	secret.Name = "pwsecret"
+	secret.Namespace = "ns-a"
+
+	mock := &test.MockClient{
+		MockList: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+			rs := &v1alpha1.RoleList{
+				Items: []v1alpha1.Role{
+					{ObjectMeta: v1.ObjectMeta{Namespace: "ns-a", Name: "r1"}, Spec: v1alpha1.RoleSpec{ForProvider: v1alpha1.RoleParameters{PasswordSecretRef: &xpv1.SecretKeySelector{SecretReference: xpv1.SecretReference{Name: "pwsecret", Namespace: "ns-a"}}}}},
+					{ObjectMeta: v1.ObjectMeta{Namespace: "ns-b", Name: "r2"}, Spec: v1alpha1.RoleSpec{ForProvider: v1alpha1.RoleParameters{PasswordSecretRef: &xpv1.SecretKeySelector{SecretReference: xpv1.SecretReference{Name: "other", Namespace: "ns-b"}}}}},
+				},
+			}
+			*list.(*v1alpha1.RoleList) = *rs
+			return nil
+		},
+	}
+
+	reqs := clusterSecretMapper(mock, secret)
+	if len(reqs) != 1 {
+		t.Fatalf("expected 1 request, got %d", len(reqs))
+	}
+	if reqs[0].Name != "r1" {
+		t.Fatalf("unexpected reconcile name: %v", reqs[0])
+	}
+}
+
+func Test_clusterSecretMapper_SecretDataChange(t *testing.T) {
+	secret := &corev1.Secret{
+		ObjectMeta: v1.ObjectMeta{Name: "pwsecret", Namespace: "ns-a"},
+		Data:       map[string][]byte{"password": []byte("old")},
+	}
+
+	mock := &test.MockClient{
+		MockList: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+			rs := &v1alpha1.RoleList{
+				Items: []v1alpha1.Role{
+					{ObjectMeta: v1.ObjectMeta{Namespace: "ns-a", Name: "r1"}, Spec: v1alpha1.RoleSpec{ForProvider: v1alpha1.RoleParameters{PasswordSecretRef: &xpv1.SecretKeySelector{SecretReference: xpv1.SecretReference{Name: "pwsecret", Namespace: "ns-a"}}}}},
+					{ObjectMeta: v1.ObjectMeta{Namespace: "ns-b", Name: "r2"}, Spec: v1alpha1.RoleSpec{ForProvider: v1alpha1.RoleParameters{PasswordSecretRef: &xpv1.SecretKeySelector{SecretReference: xpv1.SecretReference{Name: "other", Namespace: "ns-b"}}}}},
+				},
+			}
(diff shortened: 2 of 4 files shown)


fernandezcuesta:
Related: #348

pkg/controller/namespaced/postgresql/role/reconciler.go
```diff
@@ -61,6 +66,20 @@ const (
 	maxConcurrency = 5
 )
 
+func namespacedSecretMapper(kube client.Client, s *corev1.Secret) []reconcile.Request {
+	var rl namespacedv1alpha1.RoleList
+	if err := kube.List(context.Background(), &rl, client.InNamespace(s.Namespace)); err != nil {
```
chlunde: `context.Background()` discards the `ctx` the handler is given at line 118, so these Lists are not cancelled on manager shutdown. Could you pass it through?

The error is dropped too: `return nil` loses the Secret event with nothing to requeue it and no log line, so a rotation that races a cache error silently falls back to the 10 minute poll. A `MapFunc` can't return an error, so `nil` is forced — a log line is not.

These two lines and `Setup` are the only lines this PR adds that no test executes.

pkg/controller/namespaced/postgresql/role/reconciler.go
```diff
@@ -61,6 +66,20 @@ const (
...
+		return nil
+	}
+	reqs := make([]reconcile.Request, 0)
+	for _, role := range rl.Items {
+		if role.Spec.ForProvider.PasswordSecretRef != nil && role.Spec.ForProvider.PasswordSecretRef.Name == s.Name {
```
chlunde: Matching on name plus the `client.InNamespace` scope above is right here — `LocalSecretKeySelector` has no namespace field — and comparing both fields in the cluster mapper is right there. Not a missed mirror; I checked it rather than flagging it.

One thing to consider adding: `role.Spec.WriteConnectionSecretToReference != nil`. `getPassword` reports a password change only by comparing the referenced Secret against the value last published to the connection Secret, so for a Role without one the enqueue can never produce a password change. Pre-existing (#131), not something this PR introduced, but this is the line where the watch could stop scheduling work that can't go anywhere.

pkg/controller/namespaced/postgresql/role/reconciler.go
```diff
@@ -86,13 +105,25 @@ func Setup(mgr ctrl.Manager, o xpcontroller.Options) error {
...
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: maxConcurrency,
-		}).
-		Complete(r)
+		})
+
+	// Watch Secrets and enqueue Roles that reference the secret as their
+	// PasswordSecretRef so that changes to the password Secret trigger
+	// reconciliation of the Role that depends on it.
+	builder = builder.Watches(&corev1.Secret{}, handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
```
chlunde: No predicate and no field index, so every Secret event in the namespace runs a full unindexed `RoleList` — a Secret nothing references costs exactly what a matching one costs. Against a fake client that is roughly 4 ms and 4 MB of allocation per event at 1000 Roles, matching or not. `TypedEnqueueRequestsFromMapFunc` also runs the mapper on both the old and the new object of an update, and the cluster and namespaced Role controllers share one Secret informer, so a single Secret update is four scans. `mgr.GetFieldIndexer().IndexField` on `passwordSecretRef` plus `client.MatchingFields` avoids it — cheap now, awkward to retrofit.

Separately, and more important: nothing tests that this call exists. Delete it from both `Setup`s (and the now-unused `handler` import) and `go test ./...` still passes, 27 ok / 0 FAIL. `Setup` is called by no test in either tree.

pkg/controller/namespaced/postgresql/role/reconciler_test.go
```diff
@@ -83,6 +83,74 @@ func (m mockDB) GetServerVersion(ctx context.Context) (int, error) {
...
 
+func Test_namespacedSecretMapper(t *testing.T) {
+	secret := &corev1.Secret{}
+	secret.Name = "pwsecret"
+	secret.Namespace = "myns"
+
+	mock := &test.MockClient{
+		MockList: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+			rs := &v1alpha1.RoleList{
+				Items: []v1alpha1.Role{
```
chlunde: This fixture has a single Role, and it matches, so the test can't see the mapper's own predicate: replace the whole `PasswordSecretRef != nil && ...Name == s.Name` condition with `if true` and both namespaced tests still pass. The cluster test has a second, non-matching Role (`r2`/`other`), and the same mutation there fails.

Three small additions would close it:
- a second Role whose `PasswordSecretRef.Name` is something else
- a Role whose `PasswordSecretRef` is nil — drop the nil-guard today and the existing tests still pass, but that Role is a nil dereference in the informer's handler goroutine, which by default takes the process down
- an assertion on `reqs[0].Namespace`, not just `.Name` — dropping `Namespace` from the emitted request passes today, on both trees

pkg/controller/namespaced/postgresql/role/reconciler_test.go
```diff
@@ -83,6 +83,74 @@ func (m mockDB) GetServerVersion(ctx context.Context) (int, error) {
...
+	reqs := namespacedSecretMapper(mock, secret)
+	if len(reqs) != 1 {
+		t.Fatalf("expected 1 request, got %d", len(reqs))
+	}
+	if reqs[0].Name != "r1" {
+		t.Fatalf("unexpected reconcile name: %v", reqs[0])
+	}
+}
+
+func Test_namespacedSecretMapper_SecretDataChange(t *testing.T) {
```
chlunde: This is the previous test twice. The mapper reads `s.Name` and `s.Namespace` and never `s.Data`, so mutating `secret.Data["password"]` and calling again runs the identical code path with the identical inputs — it can't fail for a password change, which is what its name promises.

To earn the name it has to go through the event handler rather than calling the mapper directly, or assert something about which events reach it.

pkg/controller/cluster/postgresql/role/reconciler.go
```diff
@@ -65,6 +67,20 @@ const (
 	maxConcurrency = 5
 )
 
+func clusterSecretMapper(kube client.Client, s *corev1.Secret) []reconcile.Request {
+	var rl v1alpha1.RoleList
+	if err := kube.List(context.Background(), &rl); err != nil {
```
chlunde: Same as the namespaced mapper: `context.Background()` instead of the handler's `ctx`, and a swallowed `List` error with no log.

One difference worth noting — this List takes no options at all, so it is every Role in the cluster on every Secret event, and the more expensive of the two.

pkg/controller/cluster/postgresql/role/reconciler_test.go
```diff
@@ -80,6 +80,70 @@ func (m mockDB) GetServerVersion(ctx context.Context) (int, error) {
...
+func Test_clusterSecretMapper(t *testing.T) {
+	secret := &corev1.Secret{}
+	secret.Name = "pwsecret"
+	secret.Namespace = "ns-a"
+
+	mock := &test.MockClient{
+		MockList: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+			rs := &v1alpha1.RoleList{
+				Items: []v1alpha1.Role{
+					{ObjectMeta: v1.ObjectMeta{Namespace: "ns-a", Name: "r1"}, Spec: v1alpha1.RoleSpec{ForProvider: v1alpha1.RoleParameters{PasswordSecretRef: &xpv1.SecretKeySelector{SecretReference: xpv1.SecretReference{Name: "pwsecret", Namespace: "ns-a"}}}}},
```
chlunde: Minor: these fixtures give the Roles `Namespace: "ns-a"` / `"ns-b"`, but the cluster CRD is `scope: Cluster`, so a real cluster Role's `.Namespace` is always empty. Not wrong, but not a model of production either — and it is why dropping `Namespace` from the emitted `reconcile.Request` still passes here.

The second fixture is doing real work though: it is what makes this test catch a deleted predicate where the namespaced one doesn't.

pkg/controller/namespaced/postgresql/role/reconciler.go
chlunde: `context.Background()` discards the `ctx` the handler is given at line 118, so these Lists are not cancelled on manager shutdown. Could you pass it through?

The error is dropped too: `return nil` loses the Secret event with nothing to requeue it and no log line, so a rotation that races a cache error silently falls back to the 10 minute poll. A `MapFunc` can't return an error, so `nil` is forced — a log line is not.

These two lines and `Setup` are the only lines this PR adds that no test executes.

pkg/controller/namespaced/postgresql/role/reconciler.go
chlunde: Matching on name plus the `client.InNamespace` scope above is right here — `LocalSecretKeySelector` has no namespace field — and comparing both fields in the cluster mapper is right there. Not a missed mirror; I checked it rather than flagging it.

One thing to consider adding: `role.Spec.WriteConnectionSecretToReference != nil`. `getPassword` reports a password change only by comparing the referenced Secret against the value last published to the connection Secret, so for a Role without one the enqueue can never produce a password change. Pre-existing (#131), not something this PR introduced, but this is the line where the watch could stop scheduling work that can't go anywhere.

pkg/controller/namespaced/postgresql/role/reconciler.go
chlunde: Nothing tests that this `Watches` call exists. Delete it from both `Setup`s (and the now-unused `handler` import) and `go test ./...` still passes, 27 ok / 0 FAIL — `Setup` is called by no test in either tree.

So the feature this PR adds has no test that fails if it is removed. Not a blocker on its own, but worth knowing before it lands.

chlunde: No predicate and no field index, so every Secret event in the namespace runs a full unindexed `RoleList` — a Secret nothing references costs exactly what a matching one costs. Against a fake client that is roughly 4 ms and 4 MB of allocation per event at 1000 Roles, matching or not.

`TypedEnqueueRequestsFromMapFunc` also runs the mapper on both the old and the new object of an update, and the cluster and namespaced Role controllers share one Secret informer, so a single Secret update is four scans.

`mgr.GetFieldIndexer().IndexField` on `passwordSecretRef` plus `client.MatchingFields` avoids it — cheap now, awkward to retrofit.

pkg/controller/namespaced/postgresql/role/reconciler_test.go
chlunde: This fixture has a single Role, and it matches, so the test can't see the mapper's own predicate: replace the whole `PasswordSecretRef != nil && ...Name == s.Name` condition with `if true` and both namespaced tests still pass. The cluster test has a second, non-matching Role (`r2`/`other`), and the same mutation there fails.

Three small additions would close it:
- a second Role whose `PasswordSecretRef.Name` is something else
- a Role whose `PasswordSecretRef` is nil — drop the nil-guard today and the existing tests still pass, but that Role is a nil dereference in the informer's handler goroutine, which by default takes the process down
- an assertion on `reqs[0].Namespace`, not just `.Name` — dropping `Namespace` from the emitted request passes today, on both trees

pkg/controller/namespaced/postgresql/role/reconciler_test.go
chlunde: This is the previous test twice. The mapper reads `s.Name` and `s.Namespace` and never `s.Data`, so mutating `secret.Data["password"]` and calling again runs the identical code path with the identical inputs — it can't fail for a password change, which is what its name promises.

To earn the name it has to go through the event handler rather than calling the mapper directly, or assert something about which events reach it.

pkg/controller/cluster/postgresql/role/reconciler.go
chlunde: Same as the namespaced mapper: `context.Background()` instead of the handler's `ctx`, and a swallowed `List` error with no log.

One difference worth noting — this List takes no options at all, so it is every Role in the cluster on every Secret event, and the more expensive of the two.

pkg/controller/cluster/postgresql/role/reconciler_test.go
chlunde: Minor: these fixtures give the Roles `Namespace: "ns-a"` / `"ns-b"`, but the cluster CRD is `scope: Cluster`, so a real cluster Role's `.Namespace` is always empty. Not wrong, but not a model of production either — and it is why dropping `Namespace` from the emitted `reconcile.Request` still passes here.

The second fixture is doing real work though: it is what makes this test catch a deleted predicate where the namespaced one doesn't.

PR #410, opened 2026-07-15
fix(postgresql/database): drop optionalOldSelf from strategy immutability rule

### LINKED ISSUE #409: postgres Database CRD fails to install on older API-server CEL environments (optionalOldSelf: no matching overload for 'hasValue')
### What happened?

The generated CRDs for the postgres `Database` resource (`databases.postgresql.sql.crossplane.io` and the v2 namespaced `databases.postgresql.sql.m.crossplane.io`) fail to install on Kubernetes API servers whose CEL environment does not honor `optionalOldSelf`. The API server rejects the CRD:

```
cannot establish control of object: CustomResourceDefinition
"databases.postgresql.sql.crossplane.io" is invalid:
...x-kubernetes-validations[0].rule: compilation failed:
ERROR found no matching overload for 'hasValue' applied to 'selfType...'
```

When this happens the provider's `ProviderRevision` cannot establish control and **provider-sql fails to install** on the affected cluster.

### The offending rule

Introduced in #343, the `strategy`-immutability validation on `DatabaseParameters` is generated with `optionalOldSelf=true`:

```
!oldSelf.hasValue() || ((!has(self.strategy) && !has(oldSelf.value().strategy))
 || (has(self.strategy) && has(oldSelf.value().strategy)
     && self.strategy == oldSelf.value().strategy))
```

`optionalOldSelf` (the `oldSelf.hasValue()` / `oldSelf.value()` form) requires a newer API-server CEL environment. Where it isn't honored, `oldSelf` is not wrapped as an optional, so `hasValue()`/`value()` have no matching overload and the CEL rule fails to compile — so the whole CRD is rejected.

### Works vs. fails

Same Kubernetes minor; differs only by the API-server (EKS platform) build:

| API server | Result |
|---|---|
| `v1.31.14-eks-8f14419` | installs fine |
| `v1.31.14-eks-0690643` | CRD rejected |
| `v1.30.14` (EKS) | CRD rejected |

The EKS platform build is provider-managed with no per-cluster ETA, so "wait for the newer build" isn't actionable for affected users.

### Proposed fix

The rule enforces simple immutability of `strategy`. Because it is a **transition rule** (it references `oldSelf`), it only runs on UPDATE, where `oldSelf` is always present — so `optionalOldSelf` is unnecessary. Replacing it with a plain presence-aware transition rule compiles on all Kubernetes >= 1.25:

```
(!has(self.strategy) && !has(oldSelf.strategy)) ||
(has(self.strategy) && has(oldSelf.strategy) && self.strategy == oldSelf.strategy)
```

One thing worth confirming from the #343 authors: `optionalOldSelf` also allows a rule to run at CREATE (no `oldSelf`). The "immutable after creation" intent only needs UPDATE enforcement, so a transition rule is correct — but flagging in case the rule was meant to also reject something at CREATE.

Happy to open a PR (fix ready).

### How can we reproduce it?

Apply the generated `Database` CRD to an API server whose CEL environment does not honor `optionalOldSelf` (e.g. EKS `v1.31.14-eks-0690643` or a 1.30.x cluster) with a server-side dry run; the apply fails with `compilation failed: ERROR found no matching overload for 'hasValue'`.

ronlevy1211:

Fixes #409.

The `strategy`-immutability validation on the postgres `Database` `DatabaseParameters` (added in #343) is generated with `optionalOldSelf=true`:

```
!oldSelf.hasValue() || ((!has(self.strategy) && !has(oldSelf.value().strategy))
 || (has(self.strategy) && has(oldSelf.value().strategy)
     && self.strategy == oldSelf.value().strategy))
```

`optionalOldSelf` (the `oldSelf.hasValue()` / `oldSelf.value()` form) requires a newer API-server CEL environment. On API servers that don't honor it (e.g. older EKS platform builds such as `v1.31.14-eks-0690643`, and 1.30.x clusters), `oldSelf` is not wrapped as an optional, so `hasValue()`/`value()` have no matching overload, the CEL rule fails to compile, and the whole CRD is rejected:

```
...x-kubernetes-validations[0].rule: compilation failed:
ERROR found no matching overload for 'hasValue' applied to 'selfType...'
```

When that happens the `ProviderRevision` cannot establish control and provider-sql fails to install on the affected cluster.

Because the rule is a **transition rule** (it references `oldSelf`), it only runs on UPDATE, where `oldSelf` is always present — so `optionalOldSelf` is unnecessary. This PR replaces it with a plain presence-aware transition rule that compiles on all Kubernetes >= 1.25:

```
(!has(self.strategy) && !has(oldSelf.strategy)) ||
(has(self.strategy) && has(oldSelf.strategy) && self.strategy == oldSelf.strategy)
```

Applied to both the cluster (v1, `postgresql.sql.crossplane.io`) and namespaced (v2, `postgresql.sql.m.crossplane.io`) Database types, and regenerated the CRDs. The regenerated CRD YAML contains no `optionalOldSelf`.

**Note for the #343 authors:** `optionalOldSelf` also permits a rule to run at CREATE (where there is no `oldSelf`). The "immutable after creation" intent only needs UPDATE-time enforcement, so a transition rule is correct — but flagging in case the original rule was intended to also reject something at CREATE.

### I have:

- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

Server-side dry-run apply of the regenerated `Database` CRDs succeeds on an affected older-EKS-build cluster (`v1.31.14-eks-0690643`) where the previous `optionalOldSelf` rule failed to compile; and continues to install on newer builds. `make generate` regenerates the CRDs with the new rule and no `optionalOldSelf`.

[contribution process]: https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md

### DIFF
--- a/apis/cluster/postgresql/v1alpha1/database_types.go
+++ b/apis/cluster/postgresql/v1alpha1/database_types.go
@@ -33,7 +33,7 @@ const (
 )
 
 // DatabaseParameters are the configurable fields of a Database.
-// +kubebuilder:validation:XValidation:rule="!oldSelf.hasValue() || ((!has(self.strategy) && !has(oldSelf.value().strategy)) || (has(self.strategy) && has(oldSelf.value().strategy) && self.strategy == oldSelf.value().strategy))",message="strategy field is immutable after creation",optionalOldSelf=true
+// +kubebuilder:validation:XValidation:rule="(!has(self.strategy) && !has(oldSelf.strategy)) || (has(self.strategy) && has(oldSelf.strategy) && self.strategy == oldSelf.strategy)",message="strategy field is immutable after creation"
 type DatabaseParameters struct {
 	// The role name of the user who will own the new database, or DEFAULT to
 	// use the default (namely, the user executing the command). To create a
--- a/apis/namespaced/postgresql/v1alpha1/database_types.go
+++ b/apis/namespaced/postgresql/v1alpha1/database_types.go
@@ -34,7 +34,7 @@ const (
 )
 
 // DatabaseParameters are the configurable fields of a Database.
-// +kubebuilder:validation:XValidation:rule="!oldSelf.hasValue() || ((!has(self.strategy) && !has(oldSelf.value().strategy)) || (has(self.strategy) && has(oldSelf.value().strategy) && self.strategy == oldSelf.value().strategy))",message="strategy field is immutable after creation",optionalOldSelf=true
+// +kubebuilder:validation:XValidation:rule="(!has(self.strategy) && !has(oldSelf.strategy)) || (has(self.strategy) && has(oldSelf.strategy) && self.strategy == oldSelf.strategy)",message="strategy field is immutable after creation"
 type DatabaseParameters struct {
 	// The role name of the user who will own the new database, or DEFAULT to
 	// use the default (namely, the user executing the command). To create a
--- a/package/crds/postgresql.sql.crossplane.io_databases.yaml
+++ b/package/crds/postgresql.sql.crossplane.io_databases.yaml
@@ -217,10 +217,8 @@ spec:
                 type: object
                 x-kubernetes-validations:
                 - message: strategy field is immutable after creation
-                  optionalOldSelf: true
-                  rule: '!oldSelf.hasValue() || ((!has(self.strategy) && !has(oldSelf.value().strategy))
-                    || (has(self.strategy) && has(oldSelf.value().strategy) && self.strategy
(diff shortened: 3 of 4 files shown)


fernandezcuesta:
Well I'm not in fully in favor of removing this because in an end of life Kubernetes version doesnt work :thinking: ...

ronlevy1211:
yeah, fair enough - since these versions are on their way out anyway, it's not worth changing upstream for it. I'll close this and just keep the tweak on our side. thanks for the look!

PR #423, opened 2026-07-22
fix(mysql): add lock_wait_timeout to DSN to stop statement pile-up

### LINKED ISSUE #422: MySQL: pending ALTER USER/GRANT statements pile up and make the server unresponsive
## What happened

Under a backlog of pending `ALTER USER` (and other write) statements issued by provider-sql, a MySQL/Percona server can become unresponsive — `SHOW PROCESSLIST` fills with account-management statements stuck in `Waiting for metadata lock`, and the connection count climbs until `max_connections` is exhausted.

## Root cause

1. **The driver never issues `KILL` on context cancel.** `github.com/go-sql-driver/mysql` cancels a `context` by closing the TCP socket, **not** by sending `KILL QUERY`. This is intentional and present in every release including the latest `v1.10.0`, so it cannot be fixed by upgrading. The statement keeps running/waiting server-side.
2. **`lock_wait_timeout` default is 1 year.** A statement blocked on a metadata/ACL lock (e.g. behind a backup such as Percona XtraBackup / `FLUSH TABLES WITH READ LOCK`, or a long transaction) stays pending on the server effectively forever.
3. **The reconcile deadline (60s, crossplane-runtime default) only closes the socket** — see (1) — so the provider gives up but the server thread keeps waiting.
4. **Retries re-issue the statement.** Each reconcile opens a fresh connection (the MySQL client calls `sql.Open` per query) and issues the statement again, which also blocks.
5. **Account-management statements serialize on a single global ACL lock**, so one stuck `GRANT`/`ALTER USER` stalls *every* account-management statement server-wide.

Net effect: blocked statements and their connections accumulate until the server is unresponsive. Affects all MySQL controllers (User, Grant, Database; cluster and namespaced) since they share the client. Not specific to password rotation — any `ALTER USER`/`GRANT`/DDL can trigger it.

## Proposed fix

Append `lock_wait_timeout` (and a dial `timeout`) to the MySQL DSN so a blocked statement fails fast server-side and releases instead of piling up. PR incoming.

## Not covered

- Non-lock hangs (dead network, server wedged mid-execution) still only get socket-closed-without-`KILL`; an application-level `KILL QUERY` pattern would be needed for that class.
- The per-query `sql.Open` / unbounded connection design is a separate concern (bounded shared pool).

jabbrwcky:

Under a backlog of pending `ALTER USER`/`GRANT`/DDL statements, MySQL/Percona can become unresponsive. `go-sql-driver/mysql` cancels a context by closing the TCP socket **without** sending `KILL QUERY` (intentional; all versions incl. latest v1.10.0). With MySQL's default `lock_wait_timeout` of 1 year, a statement blocked on a metadata/ACL lock (e.g. behind a backup or long transaction) stays pending server-side even after the 60s reconcile deadline. Every retry adds another; account-management statements serialize on a single global ACL lock, so they pile up until `max_connections` is exhausted or the server wedges.

This PR appends `lock_wait_timeout=30` and dial `timeout=10s` to the MySQL DSN so blocked statements fail fast server-side and release instead of accumulating. Applies to all MySQL controllers (User, Grant, Database; cluster + namespaced) via the shared client.

Not covered / follow-ups:
- Non-lock hangs (dead network, server wedged mid-execution) still only get socket-closed-without-KILL; app-level `KILL QUERY` is the only remedy.
- Bounded shared connection pool (per-query `sql.Open`) — separate concern, see #195.
- Making `lock_wait_timeout` configurable via `ProviderConfig`.

See `docs/mysql-driver-context-cancellation.md` for the full analysis.

Fixes #422

I have:

- Updated DSN assertions; added `TestDSNParsesWithLockWaitTimeout` validating the DSN parses via the driver and carries the params.
- `go build ./...`, `go vet`, and full MySQL controller test suites pass.
- `make` and `make reviewable` pass.


jabbrwcky:
Superseded by #435 (same branch got wiped by a bad rebase, GitHub wouldn't let this PR reopen after recovery).

PR #425, opened 2026-07-23
feat: configurable, reused DB connection pool

### LINKED ISSUE #195: Provider should use connection pooling or single transaction to minimize load on database when reconciling

jabbrwcky:

Every DB client previously opened a new `*sql.DB` via `sql.Open` on every query and closed it immediately, so the provider held an effectively unbounded number of short-lived connections against each server.

This PR introduces a process-wide, DSN-keyed connection-pool cache (`pkg/clients/pool`): managed resources targeting the same server (driver + DSN) now share a single bounded `*sql.DB`. Pools apply `MaxOpenConns` / `MaxIdleConns` / `ConnMaxLifetime` / `ConnMaxIdleTime` and are evicted after going idle, so a credential rotation (new DSN) does not leak pools.

The pool is tunable per `ProviderConfig` / `ClusterProviderConfig` via a new optional `connectionPool` block, for all three databases (MySQL, PostgreSQL, MSSQL), cluster- and namespaced-scoped. Omitted fields use defaults: `maxOpenConnections=10`, `maxIdleConnections=5`, `maxConnLifetime=1h`, `maxConnIdleTime=10m`.

Notable design points for reviewers:
- The `xsql.DB` interface intentionally gains no `Close()`: the cache owns pool lifecycle, so the per-controller `Disconnect` methods stay no-ops.
- Pools are shared per `(driver, DSN)`; the config that first creates a pool for a DSN wins (in practice a DSN maps to one ProviderConfig).
- Defaults are applied per-field and on a nil `connectionPool`, so existing ProviderConfigs keep working unchanged.

See `docs/connection-pool.md` for details.

Fixes #195

I have:

- New unit tests for the pool cache (reuse per DSN, distinct pools per driver/DSN, idle-sweep eviction, active pools not swept).
- Updated client and controller unit tests for the new constructor signatures.
- `make` and `make reviewable` pass (generate is a no-op diff, lint and full `go test ./...` are green).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jabbrwcky:
Superseded by #434 (same branch got wiped by a bad rebase, GitHub wouldn't let this PR reopen after recovery).

PR #433, opened 2026-08-17
feat: Upgrade crossplane-runtime to v2.4.0

### LINKED ISSUE #392: Support for crossplane-runtime 2.3
Crossplane runtime 2.3 adds some nice features such as per-resource reconciliation support.

Bump up cr to 2.3

yeldiRium:

Fixes #392

I have:

I tried to run `make reviewable`, but there is no target. Also the readme seems to be c&p'd from the provider-helm and is out of date. Not sure what to do with that.

I ran `go test ./...`. Since this change only affects types and follows the official migration guide, I don't think functional tests against a running cluster are necessary.

Also I don't have an appropriate setup for that. Sorry about that.


fernandezcuesta:
Would you mind upgrading it to 2.4? https://github.com/crossplane/crossplane-runtime/releases/tag/v2.4.0

I'm particularly interested in the CRD optimisation mentioned in the release notes there. Other than that, the heavy lifting is already done in your PR.

yeldiRium:
> Would you mind upgrading it to 2.4? https://github.com/crossplane/crossplane-runtime/releases/tag/v2.4.0
> 
> I'm particularly interested in the CRD optimisation mentioned in the release notes there. Other than that, the heavy lifting is already done in your PR.

I was hesitant to upgrade further in fear of more breaking changes, but seems there are none in 2.4.0. I'll update the PR on monday.

yeldiRium:
@fernandezcuesta I've bumped the crossplane-runtime further to v2.4.0 and re-tested. Seems good :)

jonasz-lasut:
Just FYI to consume the performance improvements you need to introduce changes similar to this one: https://github.com/crossplane-contrib/provider-terraform/pull/320/changes#diff-36b6d20eb5aea66cf39f8d94111bd96513626ef7f61459f0d9e8e9507ded1d17

yeldiRium:
> Just FYI to consume the performance improvements you need to introduce changes similar to this one: https://github.com/crossplane-contrib/provider-terraform/pull/320/changes#diff-36b6d20eb5aea66cf39f8d94111bd96513626ef7f61459f0d9e8e9507ded1d17

thank you for the hint! that is good to know.
but I'd prefer to keep that a separate change from this PR, since it's already quite big.
is that alright with you, @fernandezcuesta?

yeldiRium:
Alright, I've added the cache configuration in a separate commit.

yeldiRium:
hi @fernandezcuesta,

not sure what to do about those e2e tests.
I deployed the current version in my own crossplane cluster and it came up fine.
Is this a known issue and maybe the check just needs a restart?

fernandezcuesta:
I guess you need this in the Makefile:

```make
CROSSPLANE_VERSION ?= v2.4.0
CROSSPLANE_CLI_VERSION ?= v2.4.0
```

Checked locally and it fails with:


```
provider-sql-3b4a9f8d4304-59cbbb5749-ft87c package-runtime crossplane-sql-provider: error: Cannot create controller manager: failed to determine if *v1.CustomResourceDefinition is namespaced: no kind is registered for the type v1.CustomResourceDefinition in scheme "pkg/runtime/scheme.go:111"
```

cmd/provider/main.go
```diff
@@ -69,18 +74,27 @@ func main() {
 	cfg, err := ctrl.GetConfig()
 	kingpin.FatalIfError(err, "Cannot get API server rest config")
 
+	scheme := runtime.NewScheme()
+	kingpin.FatalIfError(clientgoscheme.AddToScheme(scheme), "Cannot add clientgo scheme")
+	kingpin.FatalIfError(apis.AddToScheme(scheme), "Cannot add SQL APIs to scheme")
+
```
jonasz-lasut, line 81: apiextensions scheme is missing, which leads to the error that @fernandezcuesta has mentioned
```go
kingpin.FatalIfError(apiextensionsv1.AddToScheme(scheme), "Cannot register k8s apiextensions APIs to scheme")
```

fernandezcuesta: came here to suggest the same, but @jonasz-lasut is faster than light :)

jonasz-lasut:
@fernandezcuesta I don't think we can use v2.4.0 Crossplane cli until the build submodule PR is merged (on that topic: I'll try and review it this week). We'll have to stick to v2.3.x for now which used the old url

go.mod
```diff
@@ -5,16 +5,19 @@ go 1.26.1
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
 	github.com/alecthomas/kingpin/v2 v2.4.0
-	github.com/crossplane/crossplane-runtime/v2 v2.2.1
+	github.com/crossplane/crossplane-runtime/v2 v2.4.0
 	github.com/crossplane/crossplane-tools v0.0.0-20250731192036-00d407d8b7ec
```
fernandezcuesta: you may need to upgrade crossplane-tools too. Also take the chance to bump k8s deps to 0.37 (see below) and `go` to 1.26.8 please 🙏. Plus go mod tidy + make generate :)

Without the latest xp-tools the resolver files are being deleted and I suspect this is why e2e tests are failing now.

go.mod
```diff
@@ -5,16 +5,19 @@ go 1.26.1
...
+	github.com/crossplane/crossplane/apis/v2 v2.3.4
```
fernandezcuesta:
```suggestion
	github.com/crossplane/crossplane/apis/v2 v2.4.0
```

go.mod
```diff
@@ -5,16 +5,19 @@ go 1.26.1
...
 	github.com/microsoft/go-mssqldb v1.10.0
 	github.com/pkg/errors v0.9.1
-	k8s.io/api v0.35.0
-	k8s.io/apimachinery v0.35.0
+	k8s.io/api v0.35.3
+	k8s.io/apiextensions-apiserver v0.35.0
+	k8s.io/apimachinery v0.35.3
+	k8s.io/client-go v0.35.3
 	k8s.io/utils v0.0.0-20260108192941-914a6e750570
 	sigs.k8s.io/controller-runtime v0.23.1
```
fernandezcuesta:
```suggestion
	k8s.io/api v0.37.0
	k8s.io/apiextensions-apiserver v0.37.0
	k8s.io/apimachinery v0.37.0
	k8s.io/client-go v0.37.0
	k8s.io/utils v0.0.0-20260707023825-cf1189d6abe3
	sigs.k8s.io/controller-runtime v0.24.1
```

yeldiRium: unfortunately as of controller-runtime v0.24.0 the schemebuilder used in all of the register.go files is deprecated and staticcheck fails because of this. it seems to me that replacing this would require quite a few changes and I don't know enough about the controller-runtime to do that without more research.

i'd appreciate some input here, otherwise it'll take a few more days to fix the linter.

jonasz-lasut: you can check this PR in provider-kubernetes that has updated crossplane-runtime and had to bump controller-runtime as well:

https://github.com/crossplane-contrib/provider-kubernetes/pull/549/changes#diff-3600de912eef216a1c0826823bfd8a4fa300f1803328eb2a85a1a90456ca334a

yeldiRium: done, thanks for the pointer

yeldiRium:
controller-runtime is migrated to 0.24 and `scheme.NewBuilder` is replaced with `runtime.NewSchemeBuilder`.
The pipelines seem to use go 1.26.7 by default so the update to 1.26.8 was an issue - I tried to fix this by setting the patch version in the workflow configs.

fernandezcuesta:
> controller-runtime is migrated to 0.24 and `scheme.NewBuilder` is replaced with `runtime.NewSchemeBuilder`. The pipelines seem to use go 1.26.7 by default so the update to 1.26.8 was an issue - I tried to fix this by setting the patch version in the workflow configs.

yeah I forgot that in this provider we're still using static values. Ideally we should use `go-version-file: go.mod` instead and let the action infer which version to use.

chlunde:
FYI CI was failing due to a flake on GitHub connectivity, rerunning fixed it.

fernandezcuesta:
LGTM. @yeldiRium thanks a lot for your thorough work!

fernandezcuesta (approved):

fernandezcuesta:
@chlunde does it look ok to you?
I can craft a release candidate image from the branch or right after merge, so we can validate it on a real cluster.

cmd/provider/main.go
```diff
@@ -23,10 +23,14 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	_ "github.com/lib/pq"
 	_ "github.com/microsoft/go-mssqldb"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/alecthomas/kingpin/v2"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
```
chlunde, line 30: this should only be corev1 as we only need secret (tested locally).

chlunde:
@fernandezcuesta Looks good, one tiny nit (Deployment schema etc not just Secret), which increases memory use slightly but we could fix it in a followup. I think master automatically generates test builds which are pullable

fernandezcuesta:
let's merge it as is, considering the effort so far and the fact that almost all other providers blindly do it like this

PR #436, opened 2026-08-19
Allow schema-qualified type names in Grant routine args

aravinthAS:
## Description of your changes

`Grant.spec.forProvider.routines[].args` validates each argument
against `^[a-zA-Z_][a-zA-Z0-9_$]*$`, which rejects any dot. This blocks
granting `EXECUTE` on routines whose signature includes a
schema-qualified composite type -- e.g. AWS RDS's
`aws_s3.table_import_from_s3`:

```
table_import_from_s3(text, text, text, aws_commons._s3_uri_1, aws_commons._aws_credentials_1)
```

Applying such a `Grant` fails at admission with:

```
spec.forProvider.routines[0].args[3]: Invalid value: "aws_commons._s3_uri_1": should match '^[a-zA-Z_][a-zA-Z0-9_$]*$'
```

There's no way to express this grant today.

### Root cause / why this isn't just "widen the regex"

The pattern isn't only shape validation -- it's the SQL-injection guard
for this field. `quotedSignatures()` in the grant reconciler splices
each argument **unquoted** directly into the `GRANT`/`REVOKE ... ON
ROUTINE` SQL string:

```go
// Safe because the CRD restricts arguments to identifier characters:
// +kubebuilder:validation:items:Pattern:=^[a-zA-Z_][a-zA-Z0-9_$]*$
args[j] = strings.ToLower(arg)
```

So the CRD pattern is the only thing standing between this field and
arbitrary SQL. This PR widens it to allow exactly one optional
`schema.type` qualification, instead of opening up punctuation
generally:

```
^[a-zA-Z_][a-zA-Z0-9_$]*(\.[a-zA-Z_][a-zA-Z0-9_$]*)?$
```

This still anchors start/end and only permits a single well-formed
`identifier.identifier` shape -- injection payloads (`; DROP TABLE
...`, embedded quotes/parens, extra dots, etc.) still fail to match.

### Changes

- Updated the pattern on `Routine.Arguments` in both the cluster-scoped
  (`apis/cluster/postgresql/v1alpha1/grant_types.go`) and namespaced
  (`apis/namespaced/postgresql/v1alpha1/grant_types.go`) Grant APIs,
  which duplicate this type.
- Updated the safety comment above `quotedSignatures()` in both
  reconcilers to document the widened invariant.
- Regenerated CRDs via `go generate` (`package/crds/postgresql.sql.crossplane.io_grants.yaml`,
  `package/crds/postgresql.sql.m.crossplane.io_grants.yaml`) -- only
  the `routines[].args` field changed.
- Added `TestRoutineArgumentPattern` (new `grant_types_test.go` in both
  API packages) asserting the pattern accepts schema-qualified
  identifiers and still rejects malformed/injection input.
- Added `RoutineArgumentsAllowSchemaQualifiedCompositeTypes` to the
  existing `TestGrantSQL` table (both cluster and namespaced
  `reconciler_test.go`), asserting the generated GRANT/REVOKE SQL for
  `aws_s3.table_import_from_s3` with `aws_commons._s3_uri_1` /
  `aws_commons._aws_credentials_1` args.

## How has this code been tested

- `go build ./...`
- `go test $(go list ./... | grep -v /test/)` -- all packages pass
- `go vet ./...`
- `gofmt -l` on all changed files -- no output
- Manually verified `go generate` regenerated only the two Grant CRD
  files, with only the `args` pattern/description changed

I have:

- [x] Run `make reviewable` to ensure this PR is ready for review. *(ran `go build`/`go test`/`go vet`/`gofmt` and `go generate` directly; `make reviewable` itself wasn't runnable in this environment, see note below)*


apis/cluster/postgresql/v1alpha1/grant_types_test.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// routineArgumentPattern must match the +kubebuilder:validation:items:Pattern
+// marker on Routine.Arguments. It's the only thing standing between that
+// field and arbitrary SQL, since the reconciler splices each argument
+// unquoted into a GRANT/REVOKE statement -- see quotedSignatures in
+// pkg/controller/cluster/postgresql/grant/reconciler.go. Keep this in sync
+// with the marker by hand; there is no way to read kubebuilder markers back
+// out of the compiled type at runtime.
+const routineArgumentPattern = `^[a-zA-Z_][a-zA-Z0-9_$]*(\.[a-zA-Z_][a-zA-Z0-9_$]*)?$`
```
chlunde: We can create a test hel;per that validates regex from CRDs instead, to avoid copy-pasting them w/ possibly dift. I think we could remove this and the namespaced test file, and instead add another test helper for regex validations in a dedicated PR

chlunde: https://github.com/crossplane-contrib/provider-sql/issues/438 - we could do this later instead

chlunde (approved):

PR #442, opened 2026-08-26
fix(postgresql): fix Role password loss on restore when privilege comparison fails

### LINKED ISSUE #441: Role: password regenerated on restore can be permanently lost when privilege comparison fails in the same Update() call
### What happened?

The "auto-regenerate role password on restore" feature from [#348](https://github.com/crossplane-contrib/provider-sql/pull/348) (v0.16.1) can regenerate and apply a new password in Postgres, then lose it permanently if a later step in the same `Update()` call fails.

In [`Update()`](https://github.com/crossplane-contrib/provider-sql/blob/fe39828c16548e9d8d9d5b061a29526fad7f4ffc/pkg/controller/cluster/postgresql/role/reconciler.go#L333-L367) (mirrored in `pkg/controller/namespaced/postgresql/role/reconciler.go`): `shouldResetPassword()` detects a restored role (`LastPasswordChange` nil, empty secret), runs `ALTER ROLE ... PASSWORD ...`, and sets [`mg.Status.AtProvider.LastPasswordChange`](https://github.com/crossplane-contrib/provider-sql/blob/fe39828c16548e9d8d9d5b061a29526fad7f4ffc/pkg/controller/cluster/postgresql/role/reconciler.go#L359) before the function has finished. The next step, [`changedPrivs(mg.Status.AtProvider.PrivilegesAsClauses, privs)`](https://github.com/crossplane-contrib/provider-sql/blob/fe39828c16548e9d8d9d5b061a29526fad7f4ffc/pkg/controller/cluster/postgresql/role/reconciler.go#L363), errors with `errComparePrivileges` if `PrivilegesAsClauses` is empty or shorter than desired ([guard](https://github.com/crossplane-contrib/provider-sql/blob/fe39828c16548e9d8d9d5b061a29526fad7f4ffc/pkg/controller/cluster/postgresql/role/reconciler.go#L186)), which happens whenever `.status` hasn't been observed yet. `Update()` returns early with an error, so `PublishConnection()` is never called and the new password never reaches the connection secret. The generic reconciler persists `mg.Status` regardless of the error, so `LastPasswordChange` is saved anyway. `shouldResetPassword()` won't fire again (no longer nil, and nothing sets `PasswordRotationTrigger` automatically), so the role is stuck with a password that's live in Postgres but recorded nowhere.

`PrivilegesAsClauses` arrives empty/short because a plain `client.Update()` earlier in the same reconcile can return a stale `.status` view, from either crossplane-runtime's `AddFinalizer` call (runs unconditionally on any object's first-ever reconcile) or this provider's own late-init persistence. The latter is deterministic: [`lateInit()`](https://github.com/crossplane-contrib/provider-sql/blob/fe39828c16548e9d8d9d5b061a29526fad7f4ffc/pkg/controller/cluster/postgresql/role/reconciler.go#L468-L501) backfills any of `Privileges.SuperUser`, `Inherit`, `CreateDb`, `CreateRole`, `Login`, `Replication`, `BypassRls`, or `ConnectionLimit` left `nil` on the spec, and reports `ResourceLateInitialized` whenever it does. A `Role` spec that doesn't set all of these hits this path on nearly every reconcile.

### How can we reproduce it?

1. Create a `Role` with `managementPolicies` excluding `Delete`, leaving `bypassRls`/`connectionLimit`/the privilege booleans unset (so late-init has something to backfill each reconcile).
2. Wait for it to be `Ready`, then delete and immediately recreate the object (fresh `.status`, external role still exists).
3. Check the connection secret: `password` comes back empty while `status.atProvider.lastPasswordChange` is set.
4. Check events for `cannot update role: cannot compare desired and observed privileges`.

provider-sql version: `v0.16.1`
Crossplane version: `v2.2.0`

### Suggested fix

- Don't persist `LastPasswordChange` until `Update()` fully succeeds. Move the assignment to just before the function's final successful return.
- Make `changedPrivs` tolerate an empty/short `existing` instead of erroring. Treat "no prior observation yet" as "apply the full desired list."

### LINKED ISSUE #443: Postgres Password Rotation Trigger is not documented

The password rotation trigger feature introduced in #348 wasn't properly documented, and the only place so far where the behavior is explained remains the PR itself.

Add it to the examples/docs.

AGejr:

Fixes a bug in the [password auto-rotation](https://github.com/crossplane-contrib/provider-sql/pull/348) feature where `Update()` on a `Role` can regenerate and apply a new password in Postgres, then lose it permanently if a later step in the same call fails.

- `changedPrivs` now applies the full desired privilege list instead of erroring when the observed clause list is shorter than desired (this happens whenever the role's `.status` hasn't been observed yet, e.g. right after a restore)
- `Update` no longer records `LastPasswordChange` until every step has succeeded, so a later failure can't strand a password that was changed in Postgres but never recorded or published

Both fixes are mirrored in the cluster-scoped and namespaced `Role` reconcilers.

Also documents the Postgres `Role` password mechanism, which was previously undocumented: a new `docs/postgresql-role-passwords.md` covering BYO vs auto-generated passwords, generation parameters, recovery of a lost password, and rotation via `passwordRotationTrigger` (including the caveat that the trigger is inert until `lastPasswordChange` is first set). Field godoc on `passwordSecretRef` and `passwordRotationTrigger` is expanded in both API groups and links to the new page, and the `Role` examples gain `rotating-role` and `byo-password-role`.

Fixes #441
Fixes #443

I have:

Updated unit tests cover the `changedPrivs` behavior change (applying the full desired set instead of erroring). The postgres e2e suite (`cluster/local/postgresdb_functions.sh`) gains coverage for the `Role` password mechanism in both the cluster and namespaced trees: BYO password publish and update propagation, auto-generated password recovery when the connection secret is lost, rotation via `passwordRotationTrigger`, and password recovery when the `Role` object is deleted and recreated while the database role persists. Full suite passes with `go test ./...`.

### DIFF
--- a/cluster/local/postgresdb_functions.sh
+++ b/cluster/local/postgresdb_functions.sh
@@ -94,6 +94,14 @@ setup_postgresdb_tests(){
   # create DB
   "${KUBECTL}" apply -f ${projectdir}/examples/${API_TYPE}/postgresql/database.yaml
 
+  echo_step "creating bring-your-own password secret for Role"
+  # byo-password-role in role.yaml references this secret; without it that Role
+  # never becomes Ready and the readiness wait below times out.
+  byo_role_pw=$(LC_ALL=C tr -cd "A-Za-z0-9" </dev/urandom | head -c 32)
+  "${KUBECTL}" create secret generic my-role-password \
+      --namespace default --save-config \
+      --from-literal password="${byo_role_pw}"
+
   echo_step "creating PostgresDB Role resource"
   # create grant
   "${KUBECTL}" apply -f ${projectdir}/examples/${API_TYPE}/postgresql/role.yaml
@@ -395,6 +403,7 @@ delete_postgresdb_resources(){
   "${KUBECTL}" delete -f "${projectdir}/examples/${API_TYPE}/postgresql/grant.yaml"
   "${KUBECTL}" delete --ignore-not-found=true -f "${projectdir}/examples/${API_TYPE}/postgresql/database.yaml"
   "${KUBECTL}" delete -f "${projectdir}/examples/${API_TYPE}/postgresql/role.yaml"
+  "${KUBECTL}" delete --ignore-not-found=true secret my-role-password -n default
   "${KUBECTL}" delete -f "${projectdir}/examples/${API_TYPE}/postgresql/schema.yaml"
   echo "${PROVIDER_CONFIG_POSTGRES_YAML}" | "${KUBECTL}" delete -f -
 
@@ -459,6 +468,99 @@ delete_extension_test() {
   echo_step_completed
 }
 
+role_connection_password() {
+  "${KUBECTL}" get secret "$1" -n default -o jsonpath='{.data.password}' 2>/dev/null | base64 --decode
+}
+
+force_reconcile_role() {
+  "${KUBECTL}" annotate --overwrite \
+    "role.postgresql.sql.${APIGROUP_SUFFIX}crossplane.io/$1" "reconcile=$(date +%s)-${RANDOM}" > /dev/null
+}
+
+# Waits until role_connection_password for $1 is non-empty and different from $2,
+# up to ~60s. Prints nothing; returns non-zero on timeout.
+wait_role_password_changed() {
+  local secret=$1 previous=$2 current _
+  for _ in $(seq 1 30); do
+    current=$(role_connection_password "${secret}")
+    if [ -n "${current}" ] && [ "${current}" != "${previous}" ]; then
+      return 0
+    fi
+    sleep 2
+  done
+  return 1
+}
+
+check_role_passwords() {
+  # bring-your-own password: the value from my-role-password must be published to
+  # the Role's own connection secret.
+  echo_step "check Role bring-your-own password is published to the connection secret"
+  local conn_pw
+  conn_pw=$(role_connection_password byo-password-role-secret)
+  if [ "${conn_pw}" != "${byo_role_pw}" ]; then
+    echo_error "ERROR: byo-password-role connection secret password does not match my-role-password"
+  fi
+  echo_step_completed
+
+  # bring-your-own password rotation: updating my-role-password must propagate.
+  echo_step "check Role bring-your-own password update propagates to the connection secret"
+  byo_role_pw=$(LC_ALL=C tr -cd "A-Za-z0-9" </dev/urandom | head -c 32)
+  "${KUBECTL}" create secret generic my-role-password -n default \
+      --from-literal password="${byo_role_pw}" --dry-run=client -o yaml | "${KUBECTL}" apply -f -
+  force_reconcile_role byo-password-role
+  if ! wait_role_password_changed byo-password-role-secret "${conn_pw}"; then
+    echo_error "ERROR: byo-password-role connection secret not updated after my-role-password changed"
+  fi
+  echo_step_completed
+
+  # auto-generated password recovery: losing the connection secret must
+  # regenerate the password and republish it. This also exercises the restore
+  # path where Observe has not yet populated the privilege clauses.
+  echo_step "check Role auto-generated password is regenerated when the connection secret is lost"
+  local old_pw
+  old_pw=$(role_connection_password rotating-role-secret)
+  "${KUBECTL}" delete secret rotating-role-secret -n default
+  force_reconcile_role rotating-role
+  if ! wait_role_password_changed rotating-role-secret "${old_pw}"; then
+    echo_error "ERROR: rotating-role connection secret was not regenerated after deletion"
+  fi
+  echo_step_completed
+
+  # rotation trigger: bumping passwordRotationTrigger past lastPasswordChange
+  # rotates the auto-generated password.
+  echo_step "check Role passwordRotationTrigger rotates the auto-generated password"
+  local prev_pw trigger
+  prev_pw=$(role_connection_password rotating-role-secret)
+  trigger=$(date -u -d '+1 hour' +%Y-%m-%dT%H:%M:%SZ 2>/dev/null || date -u -v+1H +%Y-%m-%dT%H:%M:%SZ)
+  "${KUBECTL}" patch "role.postgresql.sql.${APIGROUP_SUFFIX}crossplane.io/rotating-role" --type merge \
+      -p "{\"spec\":{\"forProvider\":{\"passwordRotationTrigger\":\"${trigger}\"}}}"
+  if ! wait_role_password_changed rotating-role-secret "${prev_pw}"; then
+    echo_error "ERROR: rotating-role password did not rotate after bumping passwordRotationTrigger"
+  fi
+  echo_step_completed
+
+  # object restore: deleting and recreating the Role k8s object while the database
+  # role persists must recover gracefully. Dropping Delete from managementPolicies
+  # orphans the database role on delete (spec.deletionPolicy does not exist on the
+  # namespaced API). Unlike the secret-only deletion above, the recreated object
+  # starts with an empty status and no finalizer, so its first reconcile goes
+  # through the Observe-finds-it -> Update (not Create) path with privilege
+  # clauses that may not be populated yet.
+  echo_step "check Role password recovers when the Role object is deleted and recreated"
+  "${KUBECTL}" patch "role.postgresql.sql.${APIGROUP_SUFFIX}crossplane.io/rotating-role" --type merge \
+      -p '{"spec":{"managementPolicies":["Observe","Create","Update","LateInitialize"]}}'
+  "${KUBECTL}" delete "role.postgresql.sql.${APIGROUP_SUFFIX}crossplane.io/rotating-role"
+  "${KUBECTL}" apply -f "${projectdir}/examples/${API_TYPE}/postgresql/role.yaml"
+  if ! "${KUBECTL}" wait --timeout 2m --for condition=Ready \
+      "role.postgresql.sql.${APIGROUP_SUFFIX}crossplane.io/rotating-role" > /dev/null; then
+    echo_error "ERROR: rotating-role did not become Ready after Role object recreate (possible password-recovery regression)"
+  fi
+  if [ -z "$(role_connection_password rotating-role-secret)" ]; then
+    echo_error "ERROR: rotating-role connection secret empty after Role object recreate"
+  fi
+  echo_step_completed
+}
+
 integration_tests_postgres() {
   setup_postgresdb_no_tls
   setup_provider_config_postgres_no_tls
@@ -467,6 +569,7 @@ integration_tests_postgres() {
   check_observe_only_database
   check_database_owner_ref
   check_all_roles_privileges
+  check_role_passwords
   check_all_schema_privileges
   check_custom_object_privileges
   setup_extension_test
--- a/docs/postgresql-role-passwords.md
+++ b/docs/postgresql-role-passwords.md
@@ -1,0 +1,84 @@
+# PostgreSQL Role passwords
+
+`provider-sql` sets a password on every PostgreSQL `Role` it manages and publishes the current value to the connection secret named by `spec.writeConnectionSecretToRef`. This document describes where that password comes from and how to rotate it.
+
+## Providing the password
+
+There are two ways to provide the password, selected by whether `spec.forProvider.passwordSecretRef` is set.
+
+### Bring your own password
+
+Set `passwordSecretRef` to the secret key that holds the password. The provider applies that value when it creates the role and re-applies it whenever the referenced secret changes, so the password is always whatever you put in that secret.
+
+For a namespaced `Role` (`postgresql.sql.m.crossplane.io`), `passwordSecretRef` takes only `name` and `key` and always resolves in the `Role`'s own namespace; drop the `namespace` field from the example below.
+
+```yaml
+apiVersion: postgresql.sql.crossplane.io/v1alpha1
+kind: Role
+metadata:
+  name: example-role
+spec:
+  forProvider:
+    privileges:
+      login: true
+    passwordSecretRef:
+      name: example-role-password
+      namespace: default
+      key: password
+  writeConnectionSecretToRef:
+    name: example-role-secret
+    namespace: default
+```
+
+### Auto-generated password
+
+When `passwordSecretRef` is not set, the provider generates a password while creating the role and writes it to the connection secret. The generated password is a cryptographically random 27 characters drawn from `a-z`, `A-Z` and `0-9` (no symbols). Length and character set are not configurable.
+
+Once the role has a generated password the provider leaves it alone, with two exceptions:
+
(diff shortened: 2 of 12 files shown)


fernandezcuesta:
Could you please also fix #443? :pray:

AGejr:
Sure, no problem. Should I open a new pr or add the documentation alongside the changes in this pr?

fernandezcuesta:
Just add it in this same PR, I realized all sort of doc was missing while reviewing it (my fault)

AGejr:
Pushed the docs and tests.

**#443**: `b25eda8` adds `docs/postgresql-role-passwords.md` covering BYO vs auto-generated passwords, the generation parameters, recovering a lost password, and `passwordRotationTrigger` (including that it does nothing until `lastPasswordChange` is set). Expanded the `passwordSecretRef` and `passwordRotationTrigger` godoc in both API groups to link the page, and regenerated the Role CRDs so the link shows up on doc.crds.dev and the marketplace.

**e2e**: `5370c61` adds `check_role_passwords` to the postgres suite, in both the cluster and namespaced trees:

- BYO password lands in the connection secret, and updates when the source secret changes
- auto-generated password is regenerated when the connection secret is deleted
- `passwordRotationTrigger` rotates it
- deleting and recreating the Role object while the DB role still exists recovers the password

On `changedPrivs`: when the observed clause list is shorter than desired, it now applies the full desired set instead of erroring. The thinking is that without a full prior observation to diff against, applying the spec seems a safer default than an error that nothing recovers from - by the time the old path errored, the password had already been changed in Postgres and `LastPasswordChange` stamped.

AGejr:
Thanks for running this through your e2e harness @chlunde, both probes are real and I'd like to address them here.

**Future-dated `passwordRotationTrigger`** - confirmed. `shouldResetPassword` compares `trigger.After(lastPasswordChange)`, and since each rotation only stamps `lastPasswordChange = now`, a trigger in the future stays "after" it and rotates on every reconcile until the wall clock passes it. My own e2e uses the `now+1h` shape and only asserts a single rotation, so it walked straight past this.

Proposal: record the trigger the provider last acted on in `status.atProvider` (`lastAppliedPasswordRotationTrigger` or similar) and rotate only when the spec trigger differs from it, instead of comparing against `lastPasswordChange`. That makes the semantics "rotate once per distinct trigger value" whether the timestamp is past or future, and the docs/example change to match ("a value you haven't used before" rather than "later than `lastPasswordChange`"). Does that seem reasonable, or do you see a cleaner approach?

**Double rotation on recreate** - agree it looks like informer cache lag: the status write recording `lastPasswordChange` hasn't landed before the next reconcile, so `shouldResetPassword` still sees a nil `LastPasswordChange` plus an empty secret and resets again. One extra regenerate on a restore is harmless on its own, but in large-scale setups with many compositions - already reconcile-heavy - the extra `ALTER ROLE` + secret write + status update per restored role adds up, so I'd rather guard it. Cheapest fix I see is a fresh `Get` of the role status right before the reset decision so we act on the latest `lastPasswordChange`. Sound right to you, or would you handle it differently?

Will push both fixes once we've settled on the approach.

pkg/controller/cluster/postgresql/role/reconciler_test.go
```diff
@@ -896,7 +902,7 @@ func TestUpdate(t *testing.T) {
 				},
 			},
 			want: want{
-				err: errors.Wrap(errors.New(errComparePrivileges), errUpdateRole),
+				err: nil,
```
chlunde: Both halves of this PR can be reverted with the suite still green. Stamp back where master had it: `ok`. `changedPrivs` from `return desired` back to `return nil`: `ok`. The case leaves `want.c` at the zero `ExternalUpdate`, and `MockExec` returns `errBoom` only on a *mismatched* query — so issuing no query at all passes too. Control pair, same `MockExec` replaced with an unconditional `return errBoom`:

    unmutated  + fail-on-any-Exec  ->  FAIL  ("cannot update role: boom")
    return nil + fail-on-any-Exec  ->  ok    (Exec is never reached)

Could you assert the statements that were issued rather than only the error? :thinking: And a case for the moved stamp — `pwchanged` true, a later `Exec` fails, assert `mg.Status.AtProvider.LastPasswordChange == nil`. `TestUpdate`'s `want` has no status field today, so that one needs the table extended. Both trees.

cluster/local/postgresdb_functions.sh
```diff
@@ -459,6 +468,99 @@ delete_extension_test() {
...
+    echo_error "ERROR: rotating-role connection secret was not regenerated after deletion"
+  fi
+  echo_step_completed
+
+  # rotation trigger: bumping passwordRotationTrigger past lastPasswordChange
+  # rotates the auto-generated password.
+  echo_step "check Role passwordRotationTrigger rotates the auto-generated password"
+  local prev_pw trigger
+  prev_pw=$(role_connection_password rotating-role-secret)
+  trigger=$(date -u -d '+1 hour' +%Y-%m-%dT%H:%M:%SZ 2>/dev/null || date -u -v+1H +%Y-%m-%dT%H:%M:%SZ)
```
chlunde: You spotted this one yourself — the `now+1h` shape with a single-rotation assertion walks straight past the repeat. Measured at this head: three reconciles with a future trigger issue three `ALTER ROLE ... PASSWORD`.

Worth changing here even before the trigger semantics do — `date -u +%FT%TZ` once the previous rotation has settled?

docs/postgresql-role-passwords.md
```diff
@@ -0,0 +1,84 @@
...
+    privileges:
+      login: true
+    # Move past status.atProvider.lastPasswordChange to rotate.
+    passwordRotationTrigger: "2026-01-01T00:00:00Z"
+  writeConnectionSecretToRef:
+    name: example-role-secret
+    namespace: default
+```
+
+> **Caveat:** `passwordRotationTrigger` has no effect until `status.atProvider.lastPasswordChange` is set, which does not happen on creation. A role the provider created itself - connection secret populated, `lastPasswordChange` still unset - ignores the trigger. `lastPasswordChange` is first recorded when the provider changes the password, which for such a role means the recovery case above. To force that first change, clear the `password` key in the connection secret; the provider regenerates it and records `lastPasswordChange`, after which the trigger works normally.
```
chlunde: Accurate about what the provider does today, which is the problem — I would rather not document rotating the live database password as the procedure. The role's password in Postgres is replaced and everything holding the old one breaks. Once the trigger means "a value you have not used before", the caveat and this workaround both go away?

docs/postgresql-role-passwords.md
```diff
@@ -0,0 +1,84 @@
...
+    namespace: default
+```
+
+### Auto-generated password
+
+When `passwordSecretRef` is not set, the provider generates a password while creating the role and writes it to the connection secret. The generated password is a cryptographically random 27 characters drawn from `a-z`, `A-Z` and `0-9` (no symbols). Length and character set are not configurable.
+
+Once the role has a generated password the provider leaves it alone, with two exceptions:
+
+- **Recovering a lost password.** If `status.atProvider.lastPasswordChange` has never been recorded and the connection secret is missing or has no password - for example the database role was restored from a snapshot but its connection secret was not, or the secret was deleted - the provider generates a fresh password and applies it to the role, so the database and the connection secret agree again. This needs no configuration.
```
chlunde: This does name the secret being deleted, which is right — but not what it costs. For a normally-created role the live password in Postgres is replaced, and everything still holding the old one breaks; a GitOps prune is enough. Worth saying that here rather than filing it under recovery? Pre-existing (#348, v0.16.0), not caused by this PR.

pkg/controller/cluster/postgresql/role/reconciler.go
```diff
@@ -176,27 +175,27 @@ func privilegesToClauses(p v1alpha1.RolePrivilege) []string {
...
-	// desired. If it does not, then we cannot safely compare
-	// privileges.
+func changedPrivs(existing []string, desired []string) []string {
+	// If existing is shorter than desired, we don't have a full prior
+	// observation to compare against (e.g. the role's status hasn't been
+	// persisted yet). Apply the full desired set rather than treating
+	// this as an unrecoverable comparison failure.
 	if len(existing) < len(desired) {
-		return nil, errors.New(errComparePrivileges)
+		return desired
```
chlunde: Right call for the reported case. Small thing while you are in here: the loop below only compares `existing[i]` for `i` in `desired`, so anything past `len(desired)` is never looked at. Already true on master — the old length check guarded the *shorter* case, never the longer one — but there is no length check at all now, and "both slices come from `privilegesToClauses`" is all that holds it up. A line of comment saying so?

pkg/controller/cluster/postgresql/role/reconciler_test.go
chlunde: Both halves of this PR can be reverted with the suite still green. Stamp back where master had it: `ok`. `changedPrivs` from `return desired` back to `return nil`: `ok`. The case leaves `want.c` at the zero `ExternalUpdate`, and `MockExec` returns `errBoom` only on a *mismatched* query — so issuing no query at all passes too. Control pair, same `MockExec` replaced with an unconditional `return errBoom`:

    unmutated  + fail-on-any-Exec  ->  FAIL  ("cannot update role: boom")
    return nil + fail-on-any-Exec  ->  ok    (Exec is never reached)

Could you assert the statements that were issued rather than only the error? :thinking: And a case for the moved stamp — `pwchanged` true, a later `Exec` fails, assert `mg.Status.AtProvider.LastPasswordChange == nil`. `TestUpdate`'s `want` has no status field today, so that one needs the table extended. Both trees.

cluster/local/postgresdb_functions.sh
chlunde: You spotted this one yourself. `shouldResetPassword` compares `trigger.After(lastPasswordChange)`, and each rotation stamps `lastPasswordChange = now` — so a trigger an hour ahead stays *after* it for that whole hour and rotates again on every reconcile. Measured at this head: three reconciles, three `ALTER ROLE ... PASSWORD`, each one replacing the live password.

The test cannot see it because `wait_role_password_changed` returns on the first change and the step completes.

`trigger=$(date -u +%FT%TZ)` taken after the previous rotation has settled is already in the past, so it fires exactly once — and then a second poll with the password unchanged would assert the *absence* of the repeat. Worth changing here even before the trigger semantics do?

docs/postgresql-role-passwords.md
chlunde: Accurate about what the provider does today, which is the problem — I would rather not document rotating the live database password as the procedure. The role's password in Postgres is replaced and everything holding the old one breaks. Once the trigger means "a value you have not used before", the caveat and this workaround both go away?

docs/postgresql-role-passwords.md
chlunde: This does name the secret being deleted, which is right — but not what it costs. For a normally-created role the live password in Postgres is replaced, and everything still holding the old one breaks; a GitOps prune is enough. Worth saying that here rather than filing it under recovery? Pre-existing (#348, v0.16.0), not caused by this PR.

pkg/controller/cluster/postgresql/role/reconciler.go
chlunde: Right call for the reported case. Small thing while you are in here: the loop below only compares `existing[i]` for `i` in `desired`, so anything past `len(desired)` is never looked at. Already true on master — the old length check guarded the *shorter* case, never the longer one — but there is no length check at all now, and "both slices come from `privilegesToClauses`" is all that holds it up. A line of comment saying so?

pkg/controller/cluster/postgresql/role/reconciler_test.go
chlunde: This case is `SuccessNoPasswordChange`, and it passes whether or not the PR's changes are there — so nothing here pins either of them.

`changedPrivs` reverted to returning `(nil, error)` on a short `existing`: suite still `ok`. The `LastPasswordChange` stamp moved back to where master had it: still `ok`. The reason is that `want` is the zero `ExternalUpdate` with `err: nil`, and `MockExec` only returns `errBoom` on a *mismatched* query — so a run that issues **no** query at all also passes. Control pair, with `MockExec` replaced by an unconditional `return errBoom`:

    unmutated  + fail-on-any-Exec  ->  FAIL  ("cannot update role: boom")
    return nil + fail-on-any-Exec  ->  ok    (Exec is never reached)

So a later change that quietly stops issuing the UPDATE would go green here.

Could you assert the statements actually issued rather than only the error? :thinking: (the stamp-on-failure case is worth its own comment — I have left it below.)

chlunde: Second ask, separate from the query assertion above: there is no case covering the moved `LastPasswordChange` stamp on a failed update.

`pwchanged` true, a later `Exec` returns an error, then assert `mg.Status.AtProvider.LastPasswordChange == nil` — that the stamp did *not* land when the password never reached the database. That is the whole point of moving it, and nothing currently fails if it moves back.

`TestUpdate`'s `want` has no status field today, so this one needs the table extended. Both trees.

cluster/local/postgresdb_functions.sh
chlunde: You spotted this one yourself. `shouldResetPassword` compares `trigger.After(lastPasswordChange)`, and each rotation stamps `lastPasswordChange = now` — so a trigger an hour ahead stays *after* it for that whole hour and rotates again on every reconcile. Measured at this head: three reconciles, three `ALTER ROLE ... PASSWORD`, each one replacing the live password.

The test cannot see it because `wait_role_password_changed` returns on the first change and the step completes.

`trigger=$(date -u +%FT%TZ)` taken after the previous rotation has settled is already in the past, so it fires exactly once — and then a second poll with the password unchanged would assert the *absence* of the repeat. Worth changing here even before the trigger semantics do?

docs/postgresql-role-passwords.md
chlunde: Accurate about what the provider does today, which is the problem — I would rather not document rotating the live database password as the procedure. The role's password in Postgres is replaced and everything still holding the old one breaks.

To be clear this is not something this PR introduced — it is the behaviour (#348, v0.16.0) that your `lastAppliedPasswordRotationTrigger` proposal fixes. Once the trigger means "a value you have not used before", the caveat and this workaround both go away?

docs/postgresql-role-passwords.md
chlunde: This does name the secret being deleted, which is right — but not what it costs. For a normally-created role the live password in Postgres is replaced, and everything still holding the old one breaks; a GitOps prune is enough. Worth saying that here rather than filing it under recovery? Pre-existing (#348, v0.16.0), not caused by this PR.

pkg/controller/cluster/postgresql/role/reconciler.go
chlunde: Right call for the reported case. Small thing while you are in here: the loop below only compares `existing[i]` for `i` in `desired`, so anything past `len(desired)` is never looked at — an over-long `existing` silently keeps privileges that are no longer desired.

Already true on master (the old length check guarded the *shorter* case, never the longer one), and with the check gone the only thing holding it up is that both slices come from `privilegesToClauses`. A line of comment saying so?

PR #450, opened 2026-09-02
standardize kind cluster name and drop upjet as dependency

fernandezcuesta:

Keep the default kind cluster name as other providers, so we get the same UX / predictable cluster name.

Also, drop upjet as dependency which we don't use at all.

I have:


chlunde:
Nice! I have some similar fixes, would be nice to get this out to avoid conflicts. I think in general splitting PRs that touch multiple unrelated things, like the runner OS upgrade, reformatting, splitting them, go upgrade, would make it easier to get out without conflicts? And if we want to reformat " -> ' we should probably fix all in one PR

fernandezcuesta:
Yeah it's a wip of some stuff I saw we need to upgrade soonish. The reformat is my vscode extension doing some shenanigans, will revert

Makefile
```diff
@@ -92,6 +92,8 @@ POSTGRES_VERSION ?= 18
 test-integration: $(KIND) $(KUBECTL) $(CROSSPLANE_CLI) $(HELM)
 	@$(INFO) running integration tests using kind $(KIND_VERSION)
 	@KIND_NODE_IMAGE_TAG=${KIND_NODE_IMAGE_TAG} \
+	  KIND_VERSION=${KIND_VERSION} \
+	  KIND_CLUSTER_NAME=${KIND_CLUSTER_NAME} \
```
chlunde: I think this always forwards a non-empty value — `local.xpkg.mk` (included at line 56) has `KIND_CLUSTER_NAME ?= local-dev` — so the `-inttests` fallback in the script never fires and the cluster ends up named `local-dev` (`make -n test-integration` shows it). Is that the intent? Dropping this line keeps the fallback, and `KIND_CLUSTER_NAME=foo make e2e` still overrides either way.

Makefile
```diff
@@ -32,9 +32,9 @@ GO111MODULE = on
 
 # ====================================================================================
 # Setup Kubernetes tools
-KIND_NODE_IMAGE_TAG ?= v1.35.1
-KIND_VERSION ?= v0.31.0
-KUBECTL_VERSION ?= v1.35.1
+KIND_NODE_IMAGE_TAG ?= v1.36.4
```
chlunde: Is the kind/k8s bump meant to be part of this one? It's not in the description, and the last commit undid the CLI upgrade, so I wasn't sure.

cluster/local/integration_tests.sh
```diff
@@ -56,6 +56,7 @@ if [ "$skipcleanup" != true ]; then
   function cleanup {
     echo_step "Cleaning up..."
     export KUBECONFIG=
+    cleanup_tls_cert_files
```
chlunde: This makes the `.pem`/`.srl` removal run on every exit now, including `QUICK_TEST=true`, which returns before any cert is created. Deliberate? (it does fix the leak when the TLS leg fails, so I may be reading it wrong)

PR #454, opened 2026-09-08
chore: cumulative CI improvements

fernandezcuesta:

> Placeholder for cumulative CI fixes that we find while fixing other stuff.

- Bump up runners to ubuntu-26
- Decouple CI from go version, let the action read which version to install from `go.mod` directly

I have:

n/a

### DIFF
--- a/.github/workflows/publish-provider-package.yml
+++ b/.github/workflows/publish-provider-package.yml
@@ -7,10 +7,6 @@ on:
         description: "Version string to use while publishing the package (e.g. v1.0.0-alpha.1)"
         default: ""
         required: false
-      go-version:
-        description: "Go version to use if building needs to be done"
-        default: "1.26.8"
-        required: false
 
 jobs:
   publish-provider-package:
@@ -27,7 +23,7 @@ jobs:
 
   add-extensions:
     needs: publish-provider-package
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   create-tag:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
 
     steps:
       - name: Checkout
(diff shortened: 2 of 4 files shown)


.github/workflows/publish-provider-package.yml
```diff
@@ -7,10 +7,6 @@ on:
         description: "Version string to use while publishing the package (e.g. v1.0.0-alpha.1)"
         default: ""
         required: false
-      go-version:
```
chlunde, line 10: this input is gone, but the job below still passes `go-version: ${{ github.event.inputs.go-version }}` — I think that resolves to empty now. `publish-provider-non-family.yml` has `default: '1.22'`, so the release build either gets an empty version or 1.22, and go.mod needs 1.26.

That workflow doesn't read go.mod either, so we can't decouple this one the same way — should we keep passing it explicitly here, or pin it in the reusable call? Or suggest version file as an option upstream?

.github/workflows/tag.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   create-tag:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
```
chlunde, line 15: FYI: ubuntu-26.04 is still marked preview https://github.blog/changelog/2026-06-11-new-runner-images-in-public-preview/

.github/workflows/publish-provider-package.yml
chlunde, line 17: here

chlunde (changes requested):

.github/workflows/tag.yml
fernandezcuesta: Ok I will change to ubuntu-latest. Puzzled that an LTS which is almost half a year stable is still a preview...

.github/workflows/publish-provider-package.yml
fernandezcuesta: Oh you're right, I'll fix that

.github/workflows/tag.yml
fernandezcuesta: OK it was made GA now, leaving as it is

.github/workflows/publish-provider-package.yml
fernandezcuesta: done

.github/workflows/publish-provider-package.yml
fernandezcuesta: done, added a job to read the version from go.mod

PR #461, opened 2026-09-13
Fix controller-runtime logger to suppress SetLogger warning

shashankvarma499:

controller-runtime v0.22.0 emits a `log.SetLogger(...) was never called`
warning (with a full stack trace) on every reconcile when its root logger is
left unset. The provider only called `ctrl.SetLogger` in debug mode, so this
warning fired continuously in production, polluting the logs with a stack trace
per reconcile.

Set the controller-runtime logger unconditionally: use the zap logger in debug
mode and a discard logger otherwise. This keeps the previous quiet behaviour in
non-debug mode while still satisfying controller-runtime's expectation and
eliminating the per-reconcile warning.

Fixes #342

I have:

- `go build ./...` passes.
- `go vet ./cmd/provider/...` passes.
- `go test ./... -count=1` passes.
- The change is confined to `cmd/provider/main.go` (the entrypoint), which has no
  unit-test surface; behaviour is verified by running the provider and observing
  the warning is gone.


fernandezcuesta (approved):

fernandezcuesta:
Thanks @shashankvarma499 for the fix! :rocket:

Issue #465, opened 2026-09-23
Reducing privileges on a PostgreSQL Grant never revokes the removed ones


mycodeself:
### What happened?

Removing a privilege from a PostgreSQL `Grant`'s `spec.forProvider.privileges` does not revoke it on the database. The removed privilege stays granted, and the resource never converges: `Observe` correctly reports the drift, `Create` runs on every reconcile, and the extra privilege survives each time.

Two consequences:

1. **The privilege is silently retained.** The spec says the role no longer has `INSERT`, but the role still has it. Reducing privileges is a routine security operation, so the manifest and the database disagree in the unsafe direction.
2. **The resource is permanently non-`Ready`**, and writes catalog on every reconcile. With the default `--poll=10m` that is 144 `REVOKE`+`GRANT` cycles per day, forever.

The root cause is that `Create` revokes only the privileges present in the *current* spec, not the previously granted ones and not all of them:

https://github.com/crossplane-contrib/provider-sql/blob/master/pkg/controller/namespaced/postgresql/grant/reconciler.go#L432

```go
sp := strings.Join(expandedPrivileges.ToStringSlice(), ",")
...
xsql.Query{String: fmt.Sprintf("REVOKE %s ON TABLE %s FROM %s", sp, tb, ro)},
xsql.Query{String: fmt.Sprintf("GRANT %s ON TABLE %s TO %s %s", sp, tb, ro, withOption(gp.WithOption))},
```

So when the spec shrinks from `[SELECT, INSERT]` to `[SELECT]`, the emitted statements are `REVOKE SELECT` + `GRANT SELECT`. `INSERT` is never touched.

`Update` is a no-op on the assumption that `Create` fully revokes, which does not hold:

https://github.com/crossplane-contrib/provider-sql/blob/master/pkg/controller/namespaced/postgresql/grant/reconciler.go#L1091

```go
// Update is a no-op, as permissions are fully revoked and then granted in the Create function,
// inside a transaction.
```

`Observe` compares the granted privilege set for exact equality, which is correct and is what detects the drift, it is the `Create` side that cannot act on it.

This affects all object-scoped PostgreSQL grant types, since they share the pattern: `RoleTable` (L432), `RoleSequence` (L403), `RoleRoutine` (L349), `RoleColumn` (L192), and the corresponding `deleteGrantQuery` branches. Both the `cluster` and `namespaced` controllers are affected.

### How can we reproduce it?

[...]

PR #466, opened 2026-09-23
feat: add support for wildcard grants to objects support ALL statements

mycodeself:
Adds `*` as a wildcard in the PostgreSQL Grant object lists. A list whose only element is the wildcard targets every object of that type in the schema, emitting `GRANT ... ON ALL [OBJECTS] IN SCHEMA`. Supported for tables, sequences and routines (as `{name: "*"}`).

This covers what `DefaultPrivileges` cannot: `ALTER DEFAULT PRIVILEGES` only binds objects created after it is set, so a role adopted onto a populated database gets nothing on the tables already there. Use both together.

#### Decissions and trade-offs

- Explicit "*" instead of an empty list. An empty list is an invisible signal: a templating bug producing no objects would silently grant on the whole schema.
- No new field or `GrantType`. `filledInFields` only checks list length, so `tables: ["*"]` still resolves to `RoleTable` through the existing dispatch reducing the required changes.
- Wildcard grants **revoke ALL PRIVILEGES** before granting. Revoking only what is about to be granted leaves everything else in place and the resource never converges. The trade-off is that a wildcard grant is authoritative for its role, schema and object type, so two Grants on the same target will fight. Documented on the field and in the examples. The named path keeps revoking only the listed privileges, since it coexists with column level grants (but this should be reviewed as I discovered a potential bug #465).

I have:

Unit tests for the emitted SQL, dispatch and validation, in both controllers. Verified against PostgreSQL 18.6 through the provider's own query builders.


fernandezcuesta (commented):
Nice approach — reusing existing `GrantType` dispatch via the `*` sentinel is cleaner than adding new types. A few findings from review:

## P1 — Routine wildcard Observe may infinite-loop

`aclexplode(p.proacl)` returns zero rows when `proacl IS NULL` — PostgreSQL stores NULL for routines that only carry the default ACL (EXECUTE to PUBLIC). The expected-count subquery (`SELECT COUNT(*) FROM pg_proc`) counts these routines, but the granted-count subquery can't see them through `aclexplode`. This creates a permanent count mismatch, causing Create to re-run every reconcile.

**Fix:** Add `AND ap.proacl IS NOT NULL` to the expected-count subquery (and same filter on the granted-count side). After `GRANT ON ALL ROUTINES` runs, every routine in the schema gets an explicit ACL (`proacl` becomes non-NULL), so both sides agree. Routines with only default ACLs are correctly excluded from both counts.

```go
"SELECT COUNT(*) FROM pg_proc ap " +
"INNER JOIN pg_namespace an ON ap.pronamespace = an.oid " +
"WHERE an.nspname=$1 AND ap.proacl IS NOT NULL"
```

Alternative: `COALESCE(p.proacl, acldefault('f', p.proowner))` to include default-ACL routines in the explode, but the NULL filter is simpler and correct since `GRANT ON ALL ROUTINES` sets proacl on every routine.

## P2 — Missing test coverage

- No `TestCreate`/`TestObserve`/`TestDelete` reconciler-level cases for wildcard grants (only `TestGrantSQL` string assertions exist). At least one Observe case per type exercises the full mock-DB Scan path.
- No test for wildcard + `withOption: GRANT` — verifying `is_grantable` wiring.
- No test for `tables: ["*"]` with `privileges: ["ALL"]` and `serverVersion: 150000` — verifying MAINTAIN is excluded from expansion on PG<17.

## P3 — Minor nits

- `grantTarget` receives the already-computed `prefixAndQuote`/`quotedSignatures` result which is discarded on the wildcard path. Could short-circuit the caller to skip the allocation, but not worth a separate change.
- The parameter reordering in `relationGrantQuery` (old $1-$6 → new $1-$5) is covered by `TestRelationGrantQueryParameters` but not called out in the PR description. Worth a line noting the positional change for reviewers.

pkg/controller/cluster/postgresql/grant/reconciler.go
```diff
@@ -940,6 +978,36 @@ func selectRoutineGrantQuery(gp v1alpha1.GrantParameters, q *xsql.Query) error {
...
 	sp := ep.ToStringSlice()
 
+	if v1alpha1.IsWildcardRoutines(gp.Routines) {
+		// No signature to compare, so the argument formatting subquery goes
+		// too. pg_proc is counted unfiltered because ON ALL ROUTINES IN SCHEMA
+		// covers functions, aggregates, window functions and procedures alike.
+		q.String = "SELECT COUNT(*) = (" +
+			"SELECT COUNT(*) FROM pg_proc ap " +
+			"INNER JOIN pg_namespace an ON ap.pronamespace = an.oid " +
+			"WHERE an.nspname=$1" +
```
fernandezcuesta, line 988: `aclexplode(p.proacl)` returns zero rows when `proacl IS NULL`, which is the case for routines that carry only the default ACL (EXECUTE to PUBLIC). The expected-count subquery counts those routines, but the granted-count subquery (`FROM (SELECT 1 FROM pg_proc p ... aclexplode(p.proacl) ...`) cannot see them. `aclexplode` produces no rows to aggregate, so they drop out of the `COUNT`.

This creates a permanent mismatch: expected > granted, so Observe reports `ResourceExists: false` every poll and Create re-runs the `GRANT` forever.

After `GRANT ON ALL ROUTINES IN SCHEMA` runs, PostgreSQL sets an explicit ACL on every routine (`proacl` becomes non-NULL), so filtering NULLs on both sides is safe, they represent routines that haven't been touched yet.

```suggestion
		q.String = "SELECT COUNT(*) = (" +
			"SELECT COUNT(*) FROM pg_proc ap " +
			"INNER JOIN pg_namespace an ON ap.pronamespace = an.oid " +
			"WHERE an.nspname=$1 AND ap.proacl IS NOT NULL" +
```

The same filter is needed in the granted-count subquery a few lines below (line ~990, the `FROM (SELECT 1 FROM pg_proc p` block) and in the namespaced copy.

pkg/controller/cluster/postgresql/grant/reconciler_test.go
```diff
@@ -1521,6 +1523,12 @@ func TestDelete(t *testing.T) {
 	}
 }
 
+const (
+	privSelect  = "SELECT"
+	privExecute = "EXECUTE"
+	objMyTable  = "mytable"
```
fernandezcuesta, line 1599: The wildcard paths are covered by `TestGrantSQL` (SQL string assertions) and `TestRelationGrantQueryParameters` (placeholder integrity), but there are no `TestCreate`/`TestObserve`/`TestDelete` reconciler-level cases exercising the full mock-DB flow for wildcard grants. A couple of additions that would close the gap:

- One `TestObserve` case with `tables: ["*"]` and `MockScan` returning true/false that exercises the `selectAllInSchema` path through the reconciler.
- One `TestGrantSQL` case combining wildcard tables with `privileges: ["ALL"]` and `serverVersion: 150000` that verifies the MAINTAIN expansion is excluded on PG<17 even for wildcard grants.
- One `TestGrantSQL` case with `withOption: GRANT` on a wildcard that verifies `WITH GRANT OPTION` appears in the generated SQL.

pkg/controller/namespaced/postgresql/grant/reconciler.go
```diff
@@ -895,6 +933,36 @@ func selectRoutineGrantQuery(gp v1alpha1.GrantParameters, q *xsql.Query) error {
...
 	sp := ep.ToStringSlice()
 
+	if v1alpha1.IsWildcardRoutines(gp.Routines) {
+		// No signature to compare, so the argument formatting subquery goes
+		// too. pg_proc is counted unfiltered because ON ALL ROUTINES IN SCHEMA
+		// covers functions, aggregates, window functions and procedures alike.
+		q.String = "SELECT COUNT(*) = (" +
+			"SELECT COUNT(*) FROM pg_proc ap " +
+			"INNER JOIN pg_namespace an ON ap.pronamespace = an.oid " +
+			"WHERE an.nspname=$1" +
```
fernandezcuesta, line 943: Same `proacl IS NULL` issue as the cluster variant, needs the same `AND ap.proacl IS NOT NULL` filter.

```suggestion
		q.String = "SELECT COUNT(*) = (" +
			"SELECT COUNT(*) FROM pg_proc ap " +
			"INNER JOIN pg_namespace an ON ap.pronamespace = an.oid " +
			"WHERE an.nspname=$1 AND ap.proacl IS NOT NULL" +
```

fernandezcuesta (commented):
Three findings: a convergence bug in the routine wildcard observe, a test coverage gap, and a minor note on the namespaced variant.

pkg/controller/cluster/postgresql/grant/reconciler_test.go
mycodeself: done!

pkg/controller/cluster/postgresql/grant/reconciler.go
mycodeself: The mismatch is real but it is not permanent, it is what triggers `Create`. `GRANT ON ALL ROUTINES` sets `proacl` on every routine it can grant on, so both sides agree afterwards.

Adding `AND ap.proacl IS NOT NULL` breaks the case the wildcard exists for. A routine created after the `GRANT` has `proacl NULL`, so the filter drops it from the expected count and it disappears from both sides

examples/cluster/postgresql/grant.yaml
```diff
@@ -274,4 +274,65 @@ spec:
     databaseRef:
       name: db1
     foreignServers:
-      - test_foreign_server
\ No newline at end of file
+      - test_foreign_server---
```
fernandezcuesta:
```suggestion
      - test_foreign_server
---
```
I suspect this broke the tests

fernandezcuesta (changes requested):

examples/cluster/postgresql/grant.yaml
fernandezcuesta: :point_up: missing doc separator

examples/namespaced/postgresql/role.yaml
```diff
@@ -68,3 +68,18 @@ spec:
...
+kind: Role
+metadata:
+  name: all-objects-role
+  namespace: default
+spec:
+  writeConnectionSecretToRef:
+    name: all-objects-role-secret
+  forProvider:
+    privileges:
+      login: true
```
fernandezcuesta:
```suggestion
      login: true
  providerConfigRef:
    kind: ProviderConfig
    name: default
```

fernandezcuesta (changes requested):

fernandezcuesta (approved):
LGTM, thanks @mycodeself for your contribution! :rocket:

